PR #45, opened 2018-06-03
Fix and regression test for #35

### LINKED ISSUE #35: NullPointerException when registering a ("dynamic") command when auto save configuration is disabled
### Prerequisites
* [X] If this is a question/suggestion, I have already considered talking about it on [Discord](https://discord.gg/FQ5vgW3)
* [X] This issue specifically has something to do with Twitch4J
* [X] I have attempted to [look for similar issues](https://github.com/PhilippHeuer/twitch4j/issues) already

### Description
When creating a TwitchClientBuilder and including .withAutoSaveConfiguration(false), I will get a stack trace when trying to register a command in said TwitchClient's CommandHandler. The stack trace only occurs when trying to register an instance of a Command (i.e. TestCommand()), not when registering the class (TestCommand.class).

**Triggers stack trace**
```
TestCommand command = new TestCommand();
twitchClient.getCommandHandler().registerCommand(command);
```

**No stack trace**
```
twitchClient.getCommandHandler().registerCommand(TestCommand.class);
```

### Steps to Reproduce
1. Create a TwitchClient instance via the TwitchClientBuilder (making sure to use .withAutoSaveConfiguration(false) )

```
TwitchClient twitchClient = TwitchClientBuilder.init()
	.withClientId("")
	.withClientSecret("")
	.withAutoSaveConfiguration(false)
	.withCredential("") 
	.connect();
```
2. Create an instance of a class that extends the Command class
```
TestCommand command = new TestCommand();
```

3. Register the command in the TwitchClient instance's CommandHandler 
```
twitchClient.getCommandHandler().registerCommand(command);
```

**Expected behavior:** 
The Command should register with not stack trace.

**Actual behavior:**
The Command is registered to the CommandHandler and I am able to use it in Twitch chat, however I get a stack trace during the registration.

**Stacktrace (if applicable):**
https://hastebin.com/ehiwuxelar.xml

**Version affected:**
0.10.1

### Additional Information
I believe this is a bug because of the logic of the method where the exception occurs.

```
public void saveDynamicCommands() {
		try {
			// Get all Dynamic Commands
			List<Command> commandList = getCommandMap().values().stream().filter(e -> e instanceof DynamicCommand).collect(Collectors.toList());

			// Save List to File
			ObjectMapper mapper = new ObjectMapper();
			mapper.writeValue(getCommandSaveFile(), commandList);

		} catch (Exception ex) {
			ex.printStackTrace();
		}
	}
```

The exception occurs at the line ``mapper.writeValue(getCommandSaveFile(), commandList);`` (line 283). A NullPointerException is thrown because there is no command save file because saving configurations was disabled when creating the TwitchClient instance. If user's are given the option to disable saving configurations, then either the exception should be handled quietly or there should be logic to check the user's preferences before attempting to write to the file and printing the error.

It's entirely possible that I am using this library in an unintended manner, however if I understand what is happening correctly then I think this is an issue.

YaourtGG:
Fixes #35

### DIFF
--- a/src/main/java/me/philippheuer/twitch4j/message/commands/Command.java
+++ b/src/main/java/me/philippheuer/twitch4j/message/commands/Command.java
@@ -34,7 +34,7 @@ public abstract class Command {
 	/**
 	 * Get aliases of command
 	 */
-	protected String[] commandAliases;
+	protected String[] commandAliases = new String[] {};
 
 	/**
 	 * Category of the Command
--- a/src/main/java/me/philippheuer/twitch4j/message/commands/CommandHandler.java
+++ b/src/main/java/me/philippheuer/twitch4j/message/commands/CommandHandler.java
@@ -280,7 +280,11 @@ public class CommandHandler {
 
 			// Save List to File
 			ObjectMapper mapper = new ObjectMapper();
-			mapper.writeValue(getCommandSaveFile(), commandList);
+
+			File commandSaveFile = getCommandSaveFile();
+			if (commandSaveFile != null) {
+				mapper.writeValue(commandSaveFile, commandList);
+			}
 
 		} catch (Exception ex) {
 			ex.printStackTrace();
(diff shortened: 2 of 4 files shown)


YaourtGG:
Assertions are not needed because the test will fail if an exception occurs.
This is exactly the point of this test

PR #56, opened 2018-11-04
Add module feature

YaourtGG:
Modules feature is now included as written in the current master branch (04/11/2018)


YaourtGG:
Added some javadoc fixes even though those are not related (at all) to this feature

PR #57, opened 2018-11-09
Small gradle cleanup

sgdc3:
- Updated Gradle to 4.10.2
- Define the source encoding
- Use HTML5 for javadoc generation


PhilippHeuer:
Thanks a lot ;)

PR #58, opened 2018-11-11
Remove HTML 5 Javadoc generation

YaourtGG:
Builds will fail since HTML 5 use is not supported in jdk8


PhilippHeuer:
Thanks, but i upgraded the jdk so that html5 javadocs are available.

Issue #64, opened 2019-03-20
Support tags on Streams


kiru:
### Prerequisites
* [x ] If this is a question/suggestion, I have already considered discuss it on [Discord Server](https://discord.gg/FQ5vgW3)
* [x ] This issue specifically has something to do with Twitch4J
* [x ] I have attempted to [look for similar issues](https://github.com/twitch4j/twitch4j/issues) already
* [x ] I have read and understood the [contribution guidelines](https://github.com/twitch4j/twitch4j/blob/master/.github/CONTRIBUTING.md) and I accept them.

### Description

Twitch streams have a list of tags. There exists an endpoint get those ( https://dev.twitch.tv/docs/api/reference/#get-all-stream-tags ). This is not yet implemented by Twitch4J. 

See also: https://discuss.dev.twitch.tv/t/introducing-the-tags-api/19574

### Steps to Reproduce

---

**Expected behavior:** 

**Actual behavior:** 

**Stacktrace (if applicable):** 

**Version affected:**

PhilippHeuer:
SomniumPrime added those functions, so i'm closing this issue. Will be part of the next alpha release `v1.0.0-alpha.11`

PR #71, opened 2019-04-05
Implement feature request described in issue #64

SomniumPrime:
Implement feature request described in issue #64. 
Specifically, adding the list of active tags on a stream returned from the getStreams method, and adding methods for the three new tag-related endpoints.

### Prerequisites
* [X ] This pull request follows the code style of the project
* [X ] I have tested this feature

...
### Issues Fixed 
* [Issue #64 ]

...

#### Added the following functionality to TwitchHelix
* modified method getStreams - Added mapping for the tagIds list to the Stream object.
* new method getAllStreamTags - get available Twitch stream tags.
* new method getStreamTags - get tags for the specified stream.
* new method replaceStreamTags - Replace (or clear) stream tags for specified stream.
#### Derived changes
* Added StreamTag and StreamTagList classes as new domain objects
* Added helper class ObjectToJsonExpander - a custom expander for encoding body parameters in JSON
...

### Additional Information 
I could not find a way to automatically map the body parameter for the replaceStreamTags, so I solved it using a custom expander instead, as described here:
https://github.com/OpenFeign/feign#custom-param-expansion

I did not add any automated integration tests, as the tags are transient (i.e. only active for 72 hours)
I did test the following use cases through an application using the new and changed methods:
* Getting the tagIds from the Stream object
* Getting all twitch stream tags, using pagination
* Getting specific stream tags
* Getting tags specific for a channel
* replacing existing tags with one new tag
* replacing existing tags with multiple new tags
* clearing existing tags

### DIFF
--- a/rest-helix/src/main/java/com/github/twitch4j/helix/ObjectToJsonExpander.java
+++ b/rest-helix/src/main/java/com/github/twitch4j/helix/ObjectToJsonExpander.java
@@ -1,0 +1,14 @@
+package com.github.twitch4j.helix;
+
+import com.github.twitch4j.common.util.TypeConvert;
+
+import feign.Param;
+
+public class ObjectToJsonExpander implements Param.Expander{
+
+	@Override
+	public String expand(Object value) {
+		return TypeConvert.objectToJson(value);
+	}
+
+}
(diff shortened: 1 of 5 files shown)


SomniumPrime:
...targeted the wrong branch

PR #74, opened 2019-04-13
fixing types for Subcription model

Flewp:
### Prerequisites
* [ x] This pull request follows the code style of the project
* [ x] I have tested this feature

...

### Changes Proposed

* [Change 1]
Change the properties to their proper types

...

### DIFF
--- a/rest-helix/src/main/java/com/github/twitch4j/helix/domain/Subscription.java
+++ b/rest-helix/src/main/java/com/github/twitch4j/helix/domain/Subscription.java
@@ -33,10 +33,10 @@ public class Subscription {
 
     // ID of the subscribed user.
     @NonNull
-    private String userId;
+    private Integer userId;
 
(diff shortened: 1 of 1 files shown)


rest-helix/src/main/java/com/github/twitch4j/helix/domain/Subscription.java
```diff
@@ -33,10 +33,10 @@
 
     // ID of the subscribed user.
     @NonNull
-    private String userId;
+    private Integer userId;
```
PhilippHeuer: You should leave the userId as String. The helix api docs specify the type as string everywhere where the user_id is used.

Parameter | Type | Description
-- | -- | --
user_id | string | ID of the subscribed user.

PhilippHeuer:
I manually made the change to fix the type, so im closing this - thanks for the PR anyway ;)

PR #76, opened 2019-05-05
Implement raid events

sg4e:
### Prerequisites

### Changes Proposed

* Implement a `RaidEvent` for incoming raids that includes the numbers of viewers joining from the raid.

### Additional Information 

This implementation parses the [IRC raid USERNOTICE as documented here](https://dev.twitch.tv/docs/irc/tags#usernotice-twitch-tags). Tested in my chat when receiving a raid and event was properly fired with the accurate raid-viewer count.

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/events/channel/RaidEvent.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/events/channel/RaidEvent.java
@@ -1,0 +1,41 @@
+package com.github.twitch4j.chat.events.channel;
+
+import com.github.twitch4j.chat.events.AbstractChannelEvent;
+import com.github.twitch4j.common.events.domain.EventChannel;
+import com.github.twitch4j.common.events.domain.EventUser;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Value;
+
+/**
+ * This event gets called when a user receives a raid.
+ * <p>
+ * This event is <i>not</i> called when a user receives a host that is not part of a raid.
+ */
+@Value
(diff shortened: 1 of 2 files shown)


chat/src/main/java/com/github/twitch4j/chat/events/channel/RaidEvent.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * This event gets called when a user receives a raid.
+ * <p>
+ * This event is <i>not</i> called when a user receives a host that is not part of a raid.
+ */
+@Value
+@Getter
+@EqualsAndHashCode(callSuper = false)
+public class RaidEvent extends AbstractChannelEvent {
```
stachu540: `@Data` is enough to implement aboved annotations

sg4e: I used the code in the other `com.github.twitch4j.chat.events.channel` events as a model.

PhilippHeuer: `@Value` is the immutable version of `@Data` without setters obviously, it's perfectly fine like this.

PhilippHeuer:
Thanks for the PR, i will merge this in later after i tested it.

chat/src/main/java/com/github/twitch4j/chat/events/channel/RaidEvent.java
stachu540: I see

PR #79, opened 2019-05-25
Implement sub streaks

sg4e:
### Prerequisites

### Changes Proposed

* Add method to query consecutive months subscribed.

### Additional Information 
The Twitch documentation is ambiguous about what value is given for 1st sub in a streak: ["[`msg-param-streak-months`] is 0 if `msg-param-should-share-streak` is 0."](https://dev.twitch.tv/docs/irc/tags/#usernotice-twitch-tags). Since only streaks >= 2 get the option in Twitch chat to "share their streak", the `USERNOTICE` should give 0 for 1st-in-streak subs. This makes sense inside the Twitch API but may confuse some users who would expect a value of 1 in this case. This PR passes the value of `msg-param-streak-months` unchanged to the `SubscriptionEvent`.

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/events/IRCEventHandler.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/events/IRCEventHandler.java
@@ -131,9 +131,13 @@ public class IRCEventHandler {
 				if(cumulativeMonths == 0) {
 					cumulativeMonths = 1;
 				}
+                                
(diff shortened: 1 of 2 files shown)


sg4e:
Sorry, just saw your comment. Yes, it's odd because the indentation issue doesn't appear in my IDE but I see it in the code files on github. Will fix for next PR.

PR #88, opened 2019-08-08
Create stream marker

sg4e:
### Prerequisites

### Changes Proposed

* Implement the Twitch Helix [Create Stream Marker](https://dev.twitch.tv/docs/api/reference/#create-stream-marker) API endpoint.

### Additional Information 
Stream Markers are created by sending a JSON message via HTTP POST to the API endpoint. The JSON message MUST include the userId of the broadcaster and MAY include a description of the marker. It is important to escape the description, if provided, so that the POST body is valid JSON; this task is handled by the `JsonEscaper` expander.


sg4e:
Marked as draft because I thought I tested the case of providing `null` for the description, but on reviewing the code while typing the PR message, I want to test again and be sure.

sg4e:
Sorry for the extended delay on this PR. The delay was caused by a confusing issue when the optional `description` field is left out (i.e. passed as `null`). In this case, when Feign makes the `POST` body for the Create Stream Marker endpoint, it leaves the template unaltered. This results in un-replaced template text being passed as the description, i.e.

```json
{
    "user_id": "123456",
    "description": "{description}"
}
```
The best solution is to provide Feign a POJO and have the JSON encoder handle the `POST` body. This properly handles optional (null, empty) fields as well as escaping characters, proper JSON structure, etc., in a much more robust way than templating in java annotations. I've given the name *Highlight* to this POJO as many names associated with stream markers are already in use as class names in the library, and marker creation on Twitch is casually called *highlighting*.

PhilippHeuer:
Thanks, looks good ;)

PR #91, opened 2019-08-15
Allow queueing synchronous work - queue .execute() instead of exceptions

MaciejGorczyca:
### Prerequisites
* [ ] This pull request follows the code style of the project

Not sure about the code style of the project, it works and looks similar, never did any thing builder-related. Not sure if this is the best way to let t4j know about Hystrix config that we want to use.

### Changes Proposed

* Allow users to queue requests instead of throwing exception.

### Additional Information 
Currently you are able to run 10 requests at the same time; trying to run more throws exception. This is default config of Hystrix. With the changes you are able to queue requests and execute them as soon as the threadpool is freed. You are still limited to 10 requests at the same time.

Example of usage:

`TwitchClientBuilder.builder().withEnableHelix(true).withClientId(SettingsUtil.clientID).withClientSecret(SettingsUtil.clientSecret).withMaxQueueSize(Integer.MAX_VALUE).build();`

```
myList = Collections.synchronizedList(new ArrayList<>());
ExecutorService executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
for (int i = 0; i < steps; i++) {
    Thread t = new Thread(() -> {
        //data = method with API request
        //myList.addAll(data); 
    });
    executorService.submit(t);
}
```
followed by 

```executorService.shutdown();
try {
    if (!executorService.awaitTermination(60, TimeUnit.SECONDS)) {
        executorService.shutdownNow();
    }
} catch (InterruptedException ex) {
    executorService.shutdownNow();
    Thread.currentThread().interrupt();
}
return myList;
```

If we don't have 10+ threads, we should put any value higher than 10 in fixed thread pool argument to check the behavior.

### DIFF
--- a/common/src/main/java/com/github/twitch4j/common/builder/TwitchAPIBuilder.java
+++ b/common/src/main/java/com/github/twitch4j/common/builder/TwitchAPIBuilder.java
@@ -1,9 +1,11 @@
 package com.github.twitch4j.common.builder;
 
 import com.github.philippheuer.events4j.EventManager;
+import com.netflix.config.ConfigurationManager;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.experimental.Wither;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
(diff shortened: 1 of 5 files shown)


PhilippHeuer:
Looks good, i just renamed the parameter to RequestQueueSize to make it more obvious what the setting is for and moved the code that sets hystrix properties into the builders, since common doesn't actually contain the feign / hystrix dependencies.

PR #93, opened 2019-09-19
Changed long to strings to reflect docs

lclc98:
### Prerequisites
* [X] This pull request follows the code style of the project
* [X] I have tested this feature

### Changes Proposed
* Breaking changes to follow twitch docs and use strings for ids instead of longs


lclc98:
Deprecating will probably require the new method to have a name change if the change was to  a list parameter.

PhilippHeuer:
You'r right, it doesn't work since it's a list.
Then i will merge it in as is, i don't want new methods.

Issue #95, opened 2019-10-08
KrakenTeam model is missing team members (users)


tedyoung:
### Prerequisites
* [X] If this is a question/suggestion, I have already considered discuss it on [Discord Server](https://discord.gg/FQ5vgW3)
* [X] This issue specifically has something to do with Twitch4J
* [X] I have attempted to [look for similar issues](https://github.com/twitch4j/twitch4j/issues) already

### Description
The "get team" API endpoint supports an array of users who belong to that team (see https://dev.twitch.tv/docs/v5/reference/teams/#get-team for details), but the current implementation in Twitch4J does not have that information.

### To Fix

The `KrakenTeam` model needs to have a `List<KrakenUser> users` added to it to support the missing information.

**Version affected:** 
1.0.0-alpha16

### Additional Information 
I have fixed this locally for me and can submit a Pull Request.

PhilippHeuer:
Thanks for the report.
I would be happy to merge this if you create a PR :)

PhilippHeuer:
Thanks, closing this as pr #95 was merged.

PR #96, opened 2019-10-08
Adds list of Team Users to getTeam API call

tedyoung:
This adds the list of Team Users to the Team API call in the rest-kraken module.

Note that the Hystrix command execution timeout was increased to 5000 ms
as the getTeam API call can take over 3500ms to execute.

Modified the TeamServiceTest to assert that there are users in the list.
This worked locally, but kept the test with @Disabled annotation.

### Prerequisites
* [X] This pull request follows the code style of the project
* [X] I have tested this feature

### Issues Fixed 
* Issue #95 - add list of users in a team

### Changes Proposed

* Added `KrakenTeamUser` - which is a user object defined in the list of users for a team and is different from other API user objects.
* Updated `KrakenTeam` - now has a `List<KrakenTeamUser> users`


tedyoung:
Thanks, and I'm glad to see the higher timeout.

For any future PRs, should I be basing it on `develop` as well?

PhilippHeuer:
Yes that would be great.

Issue #97, opened 2019-10-11
Founders do not have `ChannelPermission.SUBSCRIBER`


sg4e:
### Prerequisites
* [x] If this is a question/suggestion, I have already considered discuss it on [Discord Server](https://discord.gg/FQ5vgW3)
* [x] This issue specifically has something to do with Twitch4J
* [x] I have attempted to [look for similar issues](https://github.com/twitch4j/twitch4j/issues) already

### Description
Twitch added a new badge today for "Founders", which are the first 10 subscribers to an affiliate or the first 25 subscribers to a partner. The Founder badge replaces their subscriber badge in Twitch chat. https://help.twitch.tv/s/article/founders-badge

Twitch4j does not recognize this badge and does not properly parse Founders as subscribers.

### Steps to Reproduce

1. Listen on any chat in which Founders are chatting for `ChannelMessageEvent`s.
2. Check `ChannelMessageEvent.getPermissions()` on any Founder's messages.

---

**Expected behavior:** 
`ChannelMessageEvent.getPermissions().contains(ChannelPermission.SUBSCRIBER)` is true.

**Actual behavior:** 
`ChannelMessageEvent.getPermissions().contains(ChannelPermission.SUBSCRIBER)` is false.

**Version affected:** 
Current develop branch (commit 798b778efd1e526da4ac79a3c643503dc30436c5 ).

Credit to [Sp0ck1](https://github.com/sp0ck1) for discovering this bug.

sg4e:
Fixed by ec0dc40510592a82b761cdc4e2fb8ef1985a1585 , which has been merged into `master`. Closing.

PR #98, opened 2019-10-11
Implement support for Twitch Founders badge

sg4e:
### Prerequisites

### Issues Fixed 
* Issue #97

### Changes Proposed

* Add a `ChannelPermission` for [Founders](https://help.twitch.tv/s/article/founders-badge)
* Give Founders `ChannelPermission.FOUNDER` and `ChannelPermission.SUBSCRIBER`.

### Additional Information 
*Founders* is a new chatter type in Twitch chat. Founders are the first 10 subscribers to an affiliate or the first 25 subscribers to a partner. They receive a special badge in chat that replaces their subscriber badge.

If a Founder unsubscribes, they will no longer have the badge on their messages when their subscription expires; however, if they resubscribe, they will regain the badge. Therefore, the presence of the badge on their messages *always* indicates that they are also a subscriber. This implementation assigns messages marked with a Founders badge as having both `ChannelPermission.FOUNDER` and `ChannelPermission.SUBSCRIBER`.

### DIFF
--- a/common/src/main/java/com/github/twitch4j/common/util/TwitchUtils.java
+++ b/common/src/main/java/com/github/twitch4j/common/util/TwitchUtils.java
@@ -45,6 +45,11 @@ public class TwitchUtils {
             if (badges.containsKey("staff")) {
                 permissionSet.add(CommandPermission.TWITCHSTAFF);
             }
+            // Founder
+            if(badges.containsKey("founder")) {
+                permissionSet.add(CommandPermission.FOUNDER);
+                permissionSet.add(CommandPermission.SUBSCRIBER);
+            }
         }
         // Moderator
         if (tags.containsKey("mod") && tags.get("mod").equals("1")) {
--- a/common/src/main/java/com/github/twitch4j/common/enums/CommandPermission.java
+++ b/common/src/main/java/com/github/twitch4j/common/enums/CommandPermission.java
@@ -22,6 +22,11 @@ public enum CommandPermission {
 	SUBSCRIBER,
 
 	/**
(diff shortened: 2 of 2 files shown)


common/src/main/java/com/github/twitch4j/common/util/TwitchUtils.java
```diff
@@ -45,6 +45,11 @@
             if (badges.containsKey("staff")) {
                 permissionSet.add(CommandPermission.TWITCHSTAFF);
             }
+            // Founder
+            if(badges.containsKey("founder")) {
+                permissionSet.add(CommandPermission.FOUNDER);
+                permissionSet.add(CommandPermission.SUBSCRIBER);
```
PhilippHeuer: I read the docs about founders and found this part:
```
Streamers that took extended breaks from streaming may have lost most of their founding
subs and may see little to none of their founders in the chat today.
Those former subscribers will receive emails that they’ve received the
Founders Badge, but there’s no guarantee that they will re-sub.

Regardless, congratulations to the new Founders. You're first forever.
```
= Having the founder badge doesn't mean someone is still subbed, just that they were first - so adding sub permissions is not accurate.

We should still detect founders that are subbed based on the irc tag (not badge) - the code a bit lower:
```
// Subscriber
if (tags.containsKey("subscriber") && tags.get("subscriber").equals("1")) {
  permissionSet.add(CommandPermission.SUBSCRIBER);
}
```

PhilippHeuer: Nevermind, you do need a active sub to equip the badge so this is fine.

PR #102, opened 2019-11-04
Initial Webhooks support

Waridley:
### Prerequisites

### Changes Proposed

* Add support for requesting and viewing Twitch webhook subscriptions.

...

### Additional Information 
This does not yet add embedded web server functionality for receiving notifications, it simply provides the methods for requesting subscriptions, and getting information about current subscriptions.

The expected use for the "type" parameter of the TwitchWebhookTopic class is to pass the topic to a lambda function that can use TypeConvert.jsonToObject() to generate the Java object and dispatch a notification event containing it. I removed the WebhookNotificationEvent class following discussion on Discord regarding removing the EventManager dependency from Helix. Perhaps it belongs in the EWS or common module instead?

One issue with providing EWS support in the future is that several of the endpoints require HTTPS. Follow and Stream Change topics seem to work without HTTPS, though, and perhaps any user seriously interested in webhooks should have an HTTPS server anyway. An example use case where an EWS would still work is deploying to a cloud provider that automatically manages SSL certificates.


Waridley:
Finally cleaned up my commits and rebased on to newest develop commit. My pubsub branch is now off of this one, though, so either this needs merged first, or I'll have to separate the webhook code from that branch before merging it.

IRCEventHandler somehow got messed up along the way... it's working again, but now converted to spaces instead of tabs for indenting. I'd try harder to fix it, but I realized `.editorconfig` sets `indent_style = space` for java files, so I'm guessing this is how you want it anyway? It seems the whole project is very inconsistent with indents. Running the problematic whitespace inspection generates tons of warnings.

PhilippHeuer:
> IRCEventHandler somehow got messed up along the way... it's working again, but now converted to spaces instead of tabs for indenting. I'd try harder to fix it, but I realized .editorconfig sets indent_style = space for java files, so I'm guessing this is how you want it anyway? It seems the whole project is very inconsistent with indents. Running the problematic whitespace inspection generates tons of warnings.

You'r right, i only added that file later so older files don't follow it yet - i should run it across all files once so that they all follow the `.editorconfig` - without that pr's will prob. always contain a ton of lines with changed whitespaces which would make it harder to review.

Waridley:
The indenting issues were actually just due to tabs and spaces being somehow mixed in the same file, so they displayed fine on my machine, but not on GitHub. Fixed now.

Waridley:
Merged webhooks branch back into my develop branch so it would let me re-open the PR

PR #109, opened 2020-04-24
Added updateTitle method to TwitchKraken

sp0ck1:
Added method to update stream titles in TwitchKraken. While Kraken is deprecated according to Twitch, this is still the only way to update stream titles via the API, and the Trello road map they provide mentions nothing about finishing the Helix API to include things like this. 

This method is tested and working :)


PhilippHeuer:
Thanks for the PR :)

PR #110, opened 2020-04-29
ScheduledThreadPoolExecutor for ClientHelper and TwitchChat fixes

boostchicken:
#### ClientHelper
* Added Scheduled Thread Pool Executor to Client Helper, this is Java standard way of doing this as opposed to sleeping a thread for every 10 seconds, this should be moved to all items.  You have the ability to give it your own thread pool since most apps will likely have one
* Custom Client Helper Thread Rate
* On enable follow or channel events, check if they already have been enabled.  Previously it would add as many duplicates as you asked it to

#### TwitchChat is much more performant, reliable, and thread-safe now.  
* Capable of multiple messages per Thread loop. Before it could process one command per 250 ms, now it will process as many commands are in the queue
* Will not drop commands if Queue gets full, will wait for messages to be sent out
* Added two new functions for channel information.  
* Removed dependency on commons-collections
* Indicated volatile variables to defeat JVM caching


PhilippHeuer:
Thanks for the PR, this is a good change. I squashed the commits and merged them to develop manually because i had to filter out a few other changes from your develop branch.

Issue #121, opened 2020-05-08
TwitchClientHelper allow more than 100 channels to be added for GoLiveEvent


condolent:
### Prerequisites
* [x] If this is a question/suggestion, I have already considered discuss it on [Discord Server](https://discord.gg/FQ5vgW3)
* [x] This issue specifically has something to do with Twitch4J
* [x] I have attempted to [look for similar issues](https://github.com/twitch4j/twitch4j/issues) already

### Description

As my application is growing I've noticed that it now has stopped working since I'm making too many requests. I'm using the `ChannelGoLiveEvent` and I need more than 100  channels to listen to. Just over 500 to be exact.

### Solutions

If I've  done `twitchClient.getClientHelper().enableStreamEventListener("channel");` more than 100 times, then split the list `listenForGoLive` and inside the Client Helper send  the request multiple times with the splitted lists, to support more than 100 channels.

### Alternatives

I've tried using the `TwitchClientHelper` class to try to replicate the function and modify it myself, but I can't seem to  make it work. However I  am sure this is due to the limits of my  Java knowledge. I'm sure anyone with more knowledge would be able to do this quite easy.

JellyBrick:
My solution (temporary fix)

```kotlin
private val eventManager = EventManager().apply {
  registerEventHandler(SimpleEventHandler())
}

...

  val twitchClientOne = buildTwitchClient()
  val twitchClientTwo = buildTwitchClient()
  eventManager.getEventHandler(SimpleEventHandler::class.java).onEvent(ChannelGoLiveEvent::class.java, eventHandler)

...

private fun buildTwitchClient(): TwitchClient {
  TwitchClientBuilder.builder()
    .withClientId(clientId)
    .withClientSecret(clientSecret)
    .withEnableHelix(true)
    .withEventManager(eventManager)
    .build()
}
```

condolent:
@JellyBrick I tried that before but must've done it incorrectly. When I tried it again by a different method it works fine. Thanks

PR #130, opened 2020-05-13
TwitchClientHelper: Allow for more than 100 channels

iProdigy:
### Changes Proposed
* Create CollectionUtils class to create chunked versions of collections
* Chunk getStream calls to allow for more than 100 channels to be tracked
* Set higher `limit` for follow listener
* Allow for bulk enable/disable stream/follow listener
* Allow for enable/disable stream/follow listener without a helix call when the channel id/name is already known
* Unrelated micro-optimization while I was poking around the util package: Use an EnumSet in TwitichUtils#getPermissionsFromTags instead of a HashSet

### Issues Fixed 
* #121 

### Additional Information
If `threadRate` is too low, 429 errors will occur. In the future, we can make the delay between requests more dynamic

### DIFF
--- a/common/src/main/java/com/github/twitch4j/common/util/CollectionUtils.java
+++ b/common/src/main/java/com/github/twitch4j/common/util/CollectionUtils.java
@@ -1,0 +1,25 @@
+package com.github.twitch4j.common.util;
+
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


PhilippHeuer:
Merged this manually too, thanks :)

PR #134, opened 2020-05-15
PubSub: Implement monetary events

iProdigy:

### Changes Proposed
* `channel-bits-events-v2`
* `channel-bits-badge-unlocks`
* `channel-commerce-events-v1`
* `channel-subscribe-events-v1`


pubsub/src/main/java/com/github/twitch4j/pubsub/domain/SubscriptionData.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * RFC 3339 timestamp of when the subscription took place
+     */
+    private String time;
+
+    /**
+     * The type of subscription plan that was purchased;
+     * "Prime" or "1000" or "2000" or "3000"
+     */
+    private String subPlan;
```
iProdigy, line 47: Should this be converted to an enum

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/SubscriptionData.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * How many months in a row has the receiving user been subscribed
+     */
+    private Integer streakMonths;
+
+    /**
+     * The event type for this subscription;
+     * "sub" or "resub" or "subgift" or "anonsubgift"
+     */
+    private String context;
```
iProdigy, line 77: Should this be converted to an enum

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/SubscriptionData.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * The login name of the user that received the subscription
+     */
+    private String recipientUserName = this.userName;
+
+    /**
+     * The display name of the user that received the subscription
+     */
+    private String recipientDisplayName = this.displayName;
```
iProdigy, line 97: In the official API, these fields are only present when the subscription is gifted. Are you fine with deviating for the sake of convenience by defaulting to the actual recipient when it is not gifted?

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/SubscriptionData.java
PhilippHeuer: Yes, we should use a enum here as we're also using one for subscriptions from the chat module `com.github.twitch4j.chat.enums.SubscriptionPlan`, maybe make a new one in the common module, then we can also switch the chat module to the common enum later on.

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/SubscriptionData.java
PhilippHeuer: Lets also use a enum here as there are just 4 possible values, maybe also in common as SubType.

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/SubscriptionData.java
PhilippHeuer: Yes, i think thats a good idea 👍

iProdigy:
My changes to ChatModerationAction in this last commit also should fix the build failing issue here: https://gitlab.com/twitch4j/twitch4j/-/jobs/578054796

PR #164, opened 2020-08-01
feat: new thread approach in client helper with exp backoff

iProdigy:

### Changes Proposed
* Use `ExponentialBackoffStrategy` in `TwitchClientHelper`
* Rework threading so that delays between tasks can be adjusted on the fly without using Thread#sleep


iProdigy:
Unfortunately not the most straightforward code (we're now recursively scheduling tasks), but it has exponential backoff and avoids some locking that was present in the previous approach so this should be more performant. Also this new approach actually allows for unlimited channel tracking (the previous approach was constrained to ~80k channels for stream events if the token was solely dedicated to that task) since the thread delay applies between calls rather than chunks of calls.

PhilippHeuer (approved):

PR #176, opened 2020-08-22
feat: add nonce and reply support to chat

### LINKED ISSUE #173: Chat (unofficial): support client-nonce and reply tags
### Description
`client-nonce` is useful for checking if a message/command was sent & to ensure the same message/event is not handled multiple times (e.g. if the user accidentally had two TwitchChat instances join the same channel, they likely don't want to handle the same event twice)

`chat_replies` is an ongoing experiment where users can read/create replies to other messages in chat

### Solutions
* Add `client-nonce` getter in `IRCMessageEvent`
* Add getter for reply-related fields in `IRCMessageEvent`
* Add ability to send a chat message with nonce/reply-id specified in `TwitchChat`

### Additional context
When receiving a reply, these tags are present:

- `reply-parent-display-name`
- `reply-parent-msg-body`
- `reply-parent-msg-id`
- `reply-parent-user-id`
- `reply-parent-user-login`

Example sending a reply:
`@client-nonce=88d627a046096ac2b3c8720c9520d73e;reply-parent-msg-id=f0f9a213-043b-400f-b425-7e5ab721aff0 PRIVMSG #streamersquare :dank reply message goes here`

Note: none of this is officially documented & could change at any time

iProdigy:

### Issues Fixed 
* Fixes #173

### Changes Proposed
* Add methods to `TwitchChat` to send messages with nonce/reply msg id/other arbitrary tags
* Add nonce & reply info to `ChannelMessageEvent` (currently not applicable to other events)
* Add convenient id & nonce getters to `IRCMessageEvent`
* Add `TwitchChat#delete(String, String)`

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
@@ -15,6 +15,7 @@ import com.github.twitch4j.common.annotation.Unofficial;
 import com.github.twitch4j.common.config.ProxyConfig;
 import com.github.twitch4j.common.util.ChatReply;
 import com.github.twitch4j.common.util.CryptoUtils;
+import com.github.twitch4j.common.util.EscapeUtils;
 import com.github.twitch4j.common.util.ExponentialBackoffStrategy;
 import com.neovisionaries.ws.client.WebSocket;
 import com.neovisionaries.ws.client.WebSocketAdapter;
@@ -607,7 +608,7 @@ public class TwitchChat implements AutoCloseable {
         StringBuilder sb = new StringBuilder();
         if (tags != null && !tags.isEmpty()) {
             sb.append('@');
-            tags.forEach((k, v) -> sb.append(k).append('=').append(escapeTagValue(v)).append(';'));
+            tags.forEach((k, v) -> sb.append(k).append('=').append(EscapeUtils.escapeTagValue(v)).append(';'));
             sb.setCharAt(sb.length() - 1, ' '); // replace last semi-colon with space
         }
         sb.append("PRIVMSG #").append(channel.toLowerCase()).append(" :").append(message);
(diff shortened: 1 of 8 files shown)


chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -574,8 +577,44 @@ public void leaveChannel(String channelName) {
...
+     * @param message the message to be sent.
+     * @param tags    the message tags (unofficial).
+     */
+    public void sendMessage(String channel, String message, @Unofficial Map<String, Object> tags) {
+        StringBuilder sb = new StringBuilder();
+        if (tags != null && !tags.isEmpty()) {
+            sb.append('@');
+            tags.forEach((k, v) -> sb.append(k).append('=').append(v).append(';'));
+            sb.setLength(sb.length() - 1); // remove last semi-colon
+            sb.append(' ');
```
iProdigy, line 612: This could be replaced by `sb.setCharAt(sb.length() - 1, ' ')`

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -574,8 +577,44 @@ public void leaveChannel(String channelName) {
...
+     *
+     * @param channel the name of the channel to send the message to.
+     * @param message the message to be sent.
+     * @param tags    the message tags (unofficial).
+     */
+    public void sendMessage(String channel, String message, @Unofficial Map<String, Object> tags) {
+        StringBuilder sb = new StringBuilder();
+        if (tags != null && !tags.isEmpty()) {
+            sb.append('@');
+            tags.forEach((k, v) -> sb.append(k).append('=').append(v).append(';'));
```
iProdigy, line 610: A more robust implementation would perform escaping on `v`

PhilippHeuer (approved):

PR #184, opened 2020-09-08
feat: allow to overwrite the default event handler

PhilippHeuer:
# Prerequisites for Code Changes

### Changes Proposed

* allows to overwrite the default EventHandler in events4j, so that ie. the ReactorEventHandler can also used for internal events based on library user choice

Related to this: https://github.com/PhilippHeuer/events4j/commit/4d3e2344fa273c64c5e9f270f1607a180f8044a0

### DIFF
--- a/twitch4j/src/main/java/com/github/twitch4j/TwitchClientBuilder.java
+++ b/twitch4j/src/main/java/com/github/twitch4j/TwitchClientBuilder.java
@@ -3,7 +3,9 @@ package com.github.twitch4j;
 import com.github.philippheuer.credentialmanager.CredentialManager;
 import com.github.philippheuer.credentialmanager.CredentialManagerBuilder;
 import com.github.philippheuer.credentialmanager.domain.OAuth2Credential;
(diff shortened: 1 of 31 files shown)


iProdigy (approved):

twitch4j/src/main/java/com/github/twitch4j/TwitchClientBuilder.java
```diff
@@ -137,6 +139,12 @@
     @With
     private EventManager eventManager = null;
 
+    /**
+     * EventManager
+     */
+    @With
+    private Class<? extends IEventHandler> defaultEventHandler = null;
```
iProdigy, line 146: Perhaps this could be replicated in the module-specific builders as well

PhilippHeuer: I added it to all builders that have a eventManager reference

iProdigy: LGTM & good doc changes 👍

PR #188, opened 2020-09-13
feat: provide channel id + name in all events

PhilippHeuer:

### Changes Proposed
* will ensure that channel id + name are present in all events
* adds 2 maps to store the mapping (channel id -> name, channel name -> id) for all joined channels)

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
@@ -119,13 +119,11 @@ public class TwitchChat implements AutoCloseable {
     /**
      * Cache: ChannelId to ChannelName
      */
-    @Getter
     protected final Map<String, String> channelIdToChannelName = new ConcurrentHashMap<>();
 
     /**
      * Cache: ChannelName to ChannelId
      */
-    @Getter
     protected final Map<String, String> channelNameToChannelId = new ConcurrentHashMap<>();
 
     /**
@@ -353,6 +351,10 @@ public class TwitchChat implements AutoCloseable {
                     channelCacheLock.lock();
                     try {
                         event.getChannelName().map(String::toLowerCase).filter(currentChannels::contains).ifPresent(name -> {
+                            // remove old name -> id mapping if present (name-change)
+                            Optional.ofNullable(channelIdToChannelName.get(event.getChannelId())).ifPresent(oldName -> channelNameToChannelId.remove(oldName));
+
+                            // store mapping
                             channelNameToChannelId.put(name, event.getChannelId());
                             channelIdToChannelName.put(event.getChannelId(), name);
                         });
@@ -788,11 +790,36 @@ public class TwitchChat implements AutoCloseable {
     }
 
     /**
-     * Returns a list of all channels currently joined (without # prefix)
+     * Returns a set of all currently joined channels (without # prefix)
      *
-     * @return List Channel Names
+     * @return a set of channel names
+     * @deprecated use getChannels() instead
      */
+    @Deprecated
     public List<String> getCurrentChannels() {
         return Collections.unmodifiableList(new ArrayList<>(currentChannels));
+    }
+
+    /**
+     * Returns a set of all currently joined channels (without # prefix)
+     *
+     * @return a set of channel names
+     */
+    public Set<String> getChannels() {
+        return Collections.unmodifiableSet(currentChannels);
+    }
+
+    /**
+     * @return the cached map used for channel id to name mapping
+     */
+    public Map<String, String> getChannelIdToChannelName() {
+        return Collections.unmodifiableMap(channelIdToChannelName);
+    }
+
+    /**
+     * @return the cached map sed for channel name to id mapping
+     */
+    public Map<String, String> getChannelNameToChannelId() {
+        return Collections.unmodifiableMap(channelNameToChannelId);
     }
 }
--- a/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
@@ -49,6 +49,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 
 @Slf4j
 public class TwitchChat implements AutoCloseable {
@@ -106,9 +107,26 @@ public class TwitchChat implements AutoCloseable {
     private volatile TMIConnectionState connectionState = TMIConnectionState.DISCONNECTED;
 
     /**
-     * Channel Cache
+     * Channel Cache Lock
      */
-    protected final Set<String> channelCache = ConcurrentHashMap.newKeySet();
+    private final ReentrantLock channelCacheLock = new ReentrantLock();
+
+    /**
+     * Current Channels
+     */
+    protected final Set<String> currentChannels = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Cache: ChannelId to ChannelName
+     */
+    @Getter
+    protected final Map<String, String> channelIdToChannelName = new ConcurrentHashMap<>();
+
+    /**
+     * Cache: ChannelName to ChannelId
+     */
+    @Getter
+    protected final Map<String, String> channelNameToChannelId = new ConcurrentHashMap<>();
 
     /**
      * IRC Message Bucket
@@ -327,6 +345,23 @@ public class TwitchChat implements AutoCloseable {
 
         // register event handler
         eventManager.onEvent(ChannelMessageEvent.class, this::onChannelMessage);
+        eventManager.onEvent(IRCMessageEvent.class, event -> {
+            // we get at least one room state event with channel name + id when we join a channel, so we cache that to provide channel id + name for all events
+            if ("ROOMSTATE".equalsIgnoreCase(event.getCommandType())) {
+                // check that channel id / name are present and that we didn't leave the channel yet
+                if (event.getChannelId() != null) {
+                    channelCacheLock.lock();
+                    try {
+                        event.getChannelName().map(String::toLowerCase).filter(currentChannels::contains).ifPresent(name -> {
+                            channelNameToChannelId.put(name, event.getChannelId());
+                            channelIdToChannelName.put(event.getChannelId(), name);
+                        });
+                    } finally {
+                        channelCacheLock.unlock();
+                    }
+                }
+            }
+        });
     }
 
     /**
@@ -422,13 +457,13 @@ public class TwitchChat implements AutoCloseable {
                     sendTextToWebSocket(String.format("nick %s", userName), true);
 
                     // Join defined channels, in case we reconnect or weren't connected yet when we called joinChannel
-                    for (String channel : channelCache) {
+                    for (String channel : currentChannels) {
                         sendCommand("join", '#' + channel);
                     }
 
                     // then join to own channel - required for sending or receiving whispers
-                    if (chatCredential != null && userName != null) {
-                        sendCommand("join", '#' + chatCredential.getUserName());
+                    if (chatCredential != null && chatCredential.getUserName() != null) {
+                        joinChannel(chatCredential.getUserName().toLowerCase());
                     } else {
                         log.warn("Chat: The whispers feature is currently not available because the provided credential does not hold information about the user. Please check the documentation on how to pass the token to the credentialManager where it will be enriched with the required information.");
                     }
@@ -470,9 +505,9 @@ public class TwitchChat implements AutoCloseable {
                                 else
                                 {
                                     try {
-                                        IRCMessageEvent event = new IRCMessageEvent(message, botOwnerIds);
+                                        IRCMessageEvent event = new IRCMessageEvent(message, channelIdToChannelName, channelNameToChannelId, botOwnerIds);
 
-                                        if(event.isValid()) {
+                                        if (event.isValid()) {
                                             eventManager.publish(event);
                                         } else {
                                             log.trace("Can't parse {}", event.getRawMessage());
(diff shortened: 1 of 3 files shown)


chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -327,6 +339,17 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
 
         // register event handler
         eventManager.onEvent(ChannelMessageEvent.class, this::onChannelMessage);
+        eventManager.onEvent(IRCMessageEvent.class, event -> {
+            // we get at least one room state event with channel name + id when we join a channel, so we cache that to provide channel id + name for all events
+            if (event.getCommandType().equals("ROOMSTATE")) {
+                if (event.getChannelId() != null && !channelIdToChannelName.containsKey(event.getChannelId())) {
```
iProdigy, line 345: I might not be so strict on a contains check here; if a name change occurs, for example, we could update the cache to be consistent with the latest name

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -108,7 +108,19 @@
...
      * Channel Cache
      */
-    protected final Set<String> channelCache = ConcurrentHashMap.newKeySet();
+    protected final Set<String> currentChannels = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Cache: ChannelId to ChannelName
+     */
+    @Getter
+    protected final Map<String, String> channelIdToChannelName = new ConcurrentHashMap();
```
iProdigy, line 117: Could add `<>` in the assignment portion to avoid the raw usage warning (same applies for the second map)

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -327,6 +339,17 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
...
         // register event handler
         eventManager.onEvent(ChannelMessageEvent.class, this::onChannelMessage);
+        eventManager.onEvent(IRCMessageEvent.class, event -> {
+            // we get at least one room state event with channel name + id when we join a channel, so we cache that to provide channel id + name for all events
+            if (event.getCommandType().equals("ROOMSTATE")) {
+                if (event.getChannelId() != null && !channelIdToChannelName.containsKey(event.getChannelId())) {
+                    channelIdToChannelName.put(event.getChannelId(), event.getChannelName().get());
+                }
+                if (event.getChannelName().isPresent() && !channelNameToChannelId.containsKey(event.getChannelName().get())) {
+                    channelNameToChannelId.put(event.getChannelName().get(), event.getChannelId());
```
iProdigy, line 349: If someone sends a `JOIN` and a `PART` shortly after but the `ROOMSTATE` arrives after the `PART` has already been issued, `currentChannels` will no longer be tracking this channel but we are still caching these mappings (that will stick around in memory until another PART is issued)

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -572,9 +595,13 @@ public void joinChannel(String channelName) {
      */
     public void leaveChannel(String channelName) {
         String lowerChannelName = channelName.toLowerCase();
-        if (channelCache.remove(lowerChannelName)) {
+        if (currentChannels.remove(lowerChannelName)) {
             sendCommand("part", "#" + lowerChannelName);
             log.debug("Leaving Channel [{}].", lowerChannelName);
+
+            // clear cache
+            channelIdToChannelName.remove(channelNameToChannelId.get(lowerChannelName));
```
iProdigy, line 603: If we do `PART` before `ROOMSTATE` occurs, name => id mapping will not exist so this will throw an NPE

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
```diff
@@ -73,6 +79,18 @@
 	 */
 	private Optional<String> payload = Optional.empty();
 
+    /**
+     * Cache: ChannelId to ChannelName
+     */
+    @ToString.Exclude
```
iProdigy, line 85: Good thinking. Should we also apply `EqualsAndHashCode.Exclude`? (same goes for the second map below)

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
```diff
@@ -105,11 +123,25 @@ public IRCMessageEvent(String rawMessage) {
      * @param rawMessage  The raw message.
      * @param botOwnerIds The bot owner ids.
      */
-	public IRCMessageEvent(String rawMessage, Collection<String> botOwnerIds) {
+	public IRCMessageEvent(String rawMessage, Map<String, String> channelIdToChannelName, Map<String, String> channelNameToChannelId, Collection<String> botOwnerIds) {
 		this.rawMessage = rawMessage;
+		this.channelIdToChannelName = channelIdToChannelName;
+		this.channelNameToChannelId = channelNameToChannelId;
```
iProdigy, line 129: Is there a long-term use for saving this to a field? Currently only seeing these maps used in the constructor

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
```diff
@@ -105,11 +123,25 @@ public IRCMessageEvent(String rawMessage) {
...
 		this.parseRawMessage();
 
+        // set channel id
+        if (tags.containsKey("room-id")) {
+            channelId = tags.get("room-id");
+        }
+
+        // provide channel id or name from cache if the event was missing one
+        if (!channelName.isPresent() && channelId != null) {
+            channelName = Optional.ofNullable(channelIdToChannelName.get(channelId));
```
iProdigy, line 140: Watch out for NPE when the map itself is null (same goes for below branch)

iProdigy:
Overall good idea to workaround events that aren't sent both channel id/name from twitch. In my bot I have a similar mapping so it's nice to provide this by default. I know this PR is still in draft stage, but I left some comments on revisions :)

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
PhilippHeuer: I originally had the code somewhere else only running if getChannelId/Name was called, but decided to do it in the constructor later - but forgot to move the properties on the class.

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
PhilippHeuer: The map itself can't be null, it's final and initialized in TwitchChat - so unless we change that i don't think we have to worry about a NPE here.

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
PhilippHeuer: I removed those properties from the class

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
PhilippHeuer: I didn't think about this case at all, i added a condition in there to only track it if the channel is still in the currentChannels list (ie. we didn't leave yet)

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
iProdigy: Yeah agreed on that point - my comment was more regarding the no args constructor that assigns null for the maps

PhilippHeuer: True, i will remove that constructor as we're not using it anywhere. It would be weird to have it in if it's not usable.

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -572,9 +594,15 @@ public void joinChannel(String channelName) {
...
-        if (channelCache.remove(lowerChannelName)) {
+        if (currentChannels.remove(lowerChannelName)) {
             sendCommand("part", "#" + lowerChannelName);
             log.debug("Leaving Channel [{}].", lowerChannelName);
+
+            // clear cache
+            if (channelNameToChannelId.containsKey(lowerChannelName)) {
+                channelIdToChannelName.remove(channelNameToChannelId.get(lowerChannelName));
+                channelNameToChannelId.remove(lowerChannelName);
+            }
```
iProdigy, line 605: Could be replaced by:
```java
String cachedId = channelNameToChannelId.remove(lowerChannelName);
if (cachedId != null) channelIdToChannelName.remove(cachedId);
```
for a marginal performance gain and perhaps greater robustness to race conditions when multiple threads are doing joins/parts simultaneously

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
iProdigy: Think there's a small logic error with the currentChannels check (we're passing a boolean rather than a String 😛), here's a possible replacement:

```java
if (event.getChannelId() != null) {
    event.getChannelName().map(String::toLowerCase).filter(currentChannels::contains).ifPresent(name -> {
        channelNameToChannelId.put(name, event.getChannelId());
        channelIdToChannelName.put(event.getChannelId(), name);
    });
}
```

Also, we may need some synchronization here in case a `PART` on the same channel occurs concurrently to the `ROOMSTATE` like so:

1. `currentChannels::contains` here is true
2. `currentChannels.remove(lowerChannelName)` in `leaveChannel` is true
3. `leaveChannel` somehow continues execution before `ifPresent` is called; as a result, the mappings have been removed from the cache
4. The `ifPresent` block is finally called and the mappings are added back to the cache

Lmk if this seems too fantastical

chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
```diff
@@ -105,11 +111,23 @@ public IRCMessageEvent(String rawMessage) {
      * @param rawMessage  The raw message.
      * @param botOwnerIds The bot owner ids.
      */
-	public IRCMessageEvent(String rawMessage, Collection<String> botOwnerIds) {
+	public IRCMessageEvent(String rawMessage, Map<String, String> channelIdToChannelName, Map<String, String> channelNameToChannelId, Collection<String> botOwnerIds) {
```
iProdigy, line 114: Could update javadocs for the constructor with the new `@param`s

iProdigy:
Can't make this comment like the others because it's not adjacent to some changed code, but:

I think when we join our own channel, we could be using the `joinChannel` method so that `currentChannels` is updated and the caches can be used for events from this channel: https://github.com/twitch4j/twitch4j/blob/8ce2a47bb2e0f77b00df6047b4e050d3155f8179/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java#L453

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -106,9 +107,26 @@
...
+
+    /**
+     * Current Channels
+     */
+    protected final Set<String> currentChannels = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Cache: ChannelId to ChannelName
+     */
+    @Getter
```
iProdigy, line 122: A bit dangerous to have this getter public (and not wrapped in `UnmodifiableMap`) - same goes for the second map

iProdigy:
Looks great! Just one last nitpick from me haha

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -327,6 +345,23 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
...
+        eventManager.onEvent(IRCMessageEvent.class, event -> {
+            // we get at least one room state event with channel name + id when we join a channel, so we cache that to provide channel id + name for all events
+            if ("ROOMSTATE".equalsIgnoreCase(event.getCommandType())) {
+                // check that channel id / name are present and that we didn't leave the channel yet
+                if (event.getChannelId() != null) {
+                    channelCacheLock.lock();
+                    try {
+                        event.getChannelName().map(String::toLowerCase).filter(currentChannels::contains).ifPresent(name -> {
+                            channelNameToChannelId.put(name, event.getChannelId());
+                            channelIdToChannelName.put(event.getChannelId(), name);
```
iProdigy, line 357: If a name change has occurred, shall we remove the old name => id mapping?

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
PhilippHeuer: wrapped both getters with a unmodifiableMap

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
PhilippHeuer: added `Optional.ofNullable(channelIdToChannelName.get(event.getChannelId())).ifPresent(oldName -> channelNameToChannelId.remove(oldName));` to account for name-changes, good point.

iProdigy: Nice, it might be marginally safer to replace the ifPresent block with

```java
String oldName = channelIdToChannelName.put(event.getChannelId(), name);
if (!name.equals(oldName)) {
    channelNameToChannelId.remove(oldName, event.getChannelId());
    channelNameToChannelId.put(name, event.getChannelId());
}
```

as this avoids an id => name mapping being missing in the case of a concurrent read

iProdigy (approved):
Nice work & thanks for bearing with my nitpicks

PR #190, opened 2020-09-13
feat: add get drops entitlements endpoint to helix

iProdigy:

### Changes Proposed
* Implement `TwitchHelix#getDropsEntitlements(String, String, String, String, String, Integer)`

### Additional Information 
https://dev.twitch.tv/docs/api/reference#get-drops-entitlements (was added under a month ago to the API)

### DIFF
--- a/rest-helix/src/main/java/com/github/twitch4j/helix/domain/DropsEntitlementList.java
+++ b/rest-helix/src/main/java/com/github/twitch4j/helix/domain/DropsEntitlementList.java
@@ -3,6 +3,7 @@ package com.github.twitch4j.helix.domain;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import lombok.AccessLevel;
 import lombok.Data;
+import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


rest-helix/src/main/java/com/github/twitch4j/helix/domain/DropsEntitlementList.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class DropsEntitlementList {
+
+    /**
+     * The entitlements
+     */
+    private List<DropsEntitlement> data;
```
PhilippHeuer, line 14: There is no getter to access this

PhilippHeuer (changes requested):
Looks good, i like the workaround you did in `getPaginationCursor` to make sure we can handle it in both cases :)

rest-helix/src/main/java/com/github/twitch4j/helix/domain/DropsEntitlementList.java
iProdigy: Good catch! Added the proper annotations to the class

rest-helix/src/main/java/com/github/twitch4j/helix/domain/DropsEntitlementList.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * If provided, is the key used to fetch the next page of data.
+     * If not provided, the current response is the last page of data available.
+     */
+    private Object pagination;
```
PhilippHeuer, line 29: Lets add `@Getter(AccessLevel.PRIVATE)` here as we have the method below to access the pagination curuser.

PhilippHeuer (approved):

PR #195, opened 2020-09-21
fix: allow multiple languages to be specified in helix getStreams

iProdigy:

### Issues Fixed 
* Could only previously request a single language because the field was a `String` instead of `List<String>` and the query parms need to be repeated (since feign will not allow `;` to be used in a single string to split up the languages as it must encode it to `%3B`)

### Changes Proposed
* Create new getStreams method that changes `String language` to `List<String> language` (also we remove `communityId` as it is no longer supported by twitch)
* Deprecate old getStreams method (and point it to the new one)

### Additional Information 
Endpoint: https://dev.twitch.tv/docs/api/reference#get-streams
Changelog: https://dev.twitch.tv/docs/change-log
> 9/10/2019: Removed communities-related data fields from Get Streams and Get Streams Metadata endpoints. These fields are no longer supported.

### DIFF
--- a/docs/content/rest-helix/streams-get.md
+++ b/docs/content/rest-helix/streams-get.md
@@ -14,34 +14,35 @@ The response has a JSON payload with a data field containing an array of stream 
 ## Method Definition
 
 ```java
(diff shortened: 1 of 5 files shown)


PhilippHeuer (approved):
LGTM

PR #210, opened 2020-10-02
feat: handle unban request pubsub topic for channels and users

iProdigy:

### Changes Proposed
* Handle unofficial `channel-unban-requests.<user-id>.<channel-id>` pubsub topic (for mods to receive unban request creations and updates in real-time)
* Handle unofficial `user-unban-requests.<user-id>.<channel-id>` pubsub topic (for banned users to monitor the status of their unban request)

### DIFF
--- a/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/UpdatedUnbanRequest.java
+++ b/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/UpdatedUnbanRequest.java
@@ -1,0 +1,28 @@
+package com.github.twitch4j.pubsub.domain;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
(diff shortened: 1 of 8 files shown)


pubsub/src/main/java/com/github/twitch4j/pubsub/domain/UpdatedUnbanRequest.java
```diff
@@ -0,0 +1,28 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class UpdatedUnbanRequest extends UnbanRequest {
+
+    private String resolverId;
+
+    private String resolverLogin;
+
+    private String resolverMessage;
+
+    private String status; // e.g. "CANCELED" or "APPROVED" or "DENIED"
```
PhilippHeuer, line 26: I think normally a enum would be a better choice here as we only have 3 possible values, but since this topic is not official / undocumented a string is fine too as it's more flexible in case they introduce new / change status codes.

PhilippHeuer (approved):

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/UpdatedUnbanRequest.java
iProdigy: Agreed, that was my primary reasoning for leaving it as a string

PR #213, opened 2020-10-02
fix: default event handler

PhilippHeuer:

### Changes Proposed

Require a fully configured EventManager (and just check if it's usable by twitch4j) or fully initialize a new one.
It's better to clearly split this up and not touch a eventManager instance that might be used by other libraries too.

I'm keeping this as draft to wait for the next events4j release -> https://github.com/PhilippHeuer/events4j/pull/7

### DIFF
--- a/common/src/main/java/com/github/twitch4j/common/util/EventManagerUtils.java
+++ b/common/src/main/java/com/github/twitch4j/common/util/EventManagerUtils.java
@@ -13,7 +13,9 @@ public class EventManagerUtils {
      */
     public static EventManager validateOrInitializeEventManager(EventManager eventManager, @SuppressWarnings("rawtypes") Class defaultEventHandler) {
         if (eventManager == null) {
-            return initializeEventManager(defaultEventHandler);
+            EventManager newEM = initializeEventManager(defaultEventHandler);
+            validateEventManager(newEM);
+            return newEM;
         } else {
             validateEventManager(eventManager);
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/github/twitch4j/common/util/EventManagerUtils.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Validates the provided EventManager or initializes a new one
+     *
+     * @param eventManager EventManager
+     * @param defaultEventHandler The default eventHandler class
+     * @return EventManager
+     */
+    public static EventManager validateOrInitializeEventManager(EventManager eventManager, @SuppressWarnings("rawtypes") Class defaultEventHandler) {
+        if (eventManager == null) {
+            return initializeEventManager(defaultEventHandler);
```
iProdigy, line 16: Should we also throw a RuntimeException if someone does `.withDefaultEventHandler(null)` and doesn't set a nonnull value for `eventManager`?

iProdigy (approved):

common/src/main/java/com/github/twitch4j/common/util/EventManagerUtils.java
PhilippHeuer: Resolved this by also calling the validate method in case we initialize the EventManager in the library.

Issue #189, opened 2020-09-13
Chat: Fire specialized event on USERSTATE


iProdigy:
### Prerequisites
* [x] If this is a question/suggestion, I have already considered discuss it on [Discord Server](https://discord.gg/FQ5vgW3)
* [x] This issue specifically has something to do with Twitch4J
* [x] I have attempted to [look for similar issues](https://github.com/twitch4j/twitch4j/issues) already

### Description
There is no specialized event for `USERSTATE`. To track this, one needs to use `IRCMessageEvent`. This event provides useful information about the authenticated user when they join a channel or send a message (e.g. are they a mod? how many months have they been subbed for?)

### Solutions
Fire a specialized event on `USERSTATE` from `IRCEventHandler`. Could include some standard helpers like isModerator and other badge features.

### Alternatives
Could also consider an event for `GLOBALUSERSTATE` (the information is less useful, but it's an indicator of a successful login)

### Additional context
https://dev.twitch.tv/docs/irc/tags#userstate-twitch-tags

PR #216, opened 2020-10-04
feat: fire specialized event on USERSTATE

tadite:

### Issues Fixed 
* [Issue #189]

### Changes Proposed

* [Implement `UserStateEvent` with badges/badgeInfo, so user of event can retrieve all needed info, and helper methods for some common cases]
* [Add USERSTATE event handler in `IRCEventHandler`]

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/events/channel/UserStateEvent.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/events/channel/UserStateEvent.java
@@ -1,0 +1,131 @@
+package com.github.twitch4j.chat.events.channel;
+
+import com.github.twitch4j.chat.events.AbstractChannelEvent;
+import com.github.twitch4j.common.events.domain.EventChannel;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Value;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.OptionalInt;
+
+/**
+ * This event gets called when a user joins a channel or sends a PRIVMSG to a channel.
+ */
+@Value
+@Getter
+@EqualsAndHashCode(callSuper = false)
+public class UserStateEvent extends AbstractChannelEvent {
+
+    /**
+     * Hexadecimal RGB color code, or null if it is never set.
+     */
+    private String color;
+
+    /**
+     * The user's display name, or null if it is never set.
+     */
+    private String displayName;
+
+    /**
+     * List of emote sets.
+     */
+    private List<String> emoteSets;
+
+    /**
+     * Badges map of tags to versions.
+     */
+    private Map<String, String> badges;
+
+    /**
+     * Metadata related to the chat badges.
+     */
+    private Map<String, String> badgeInfo;
+
+    /**
+     * Event constructor
+     *
+     * @param channel     The channel that this event originates from.
+     * @param displayName The display version of the user name
+     * @param color       Hexadecimal RGB color code of the user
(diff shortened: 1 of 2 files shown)


chat/src/main/java/com/github/twitch4j/chat/events/channel/UserStateEvent.java
```diff
@@ -0,0 +1,131 @@
...
+     * Event constructor
+     *
+     * @param channel     The channel that this event originates from.
+     * @param displayName The display version of the user name
+     * @param color       Hexadecimal RGB color code of the user
+     * @param emoteSets   The emote sets the user is entitled to use.
+     * @param badges      Chat badges of the user and the version of each badge
+     * @param badgeInfo   Metadata related to the user chat badges
+     */
+    public UserStateEvent(EventChannel channel, String displayName, String color,
```
PhilippHeuer, line 57: Can you change the constructor to take a `IRCMessageEvent event` as argument and only store that reference in the class?

chat/src/main/java/com/github/twitch4j/chat/events/channel/UserStateEvent.java
```diff
@@ -0,0 +1,131 @@
...
+                          List<String> emoteSets, Map<String, String> badges, Map<String, String> badgeInfo) {
+        super(channel);
+        this.displayName = displayName;
+        this.color = color;
+        this.emoteSets = Collections.unmodifiableList(emoteSets);
+        this.badges = Collections.unmodifiableMap(badges);
+        this.badgeInfo = Collections.unmodifiableMap(badgeInfo);
+    }
+
+    /**
```
PhilippHeuer, line 67: All getters below can then directly get the information from the IRCMessageEvent - and in case anyone wants to access something else they will be able to directly get that information from the IRCMessageEvent.

PhilippHeuer (changes requested):
Thanks for your contribution :)

The pr looks good overall but we generally prefer to store the references to the origin of the event to allow access in case our users need information that's not available from the getters on the event.

chat/src/main/java/com/github/twitch4j/chat/events/channel/UserStateEvent.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return true if user is subscribed, false otherwise
+     */
+    public boolean isSubscriber() {
+        return this.badges.containsKey("subscriber");
```
iProdigy, line 71: In addition to delegating to `IRCMessageEvent`, these badge lookups could be done via `clientPermissions` like so: `messageEvent.getClientPermissions().contains(CommandPermission.SUBSCRIBER)`

tadite: Is that ok to add new items in `CommandPermission` enum and parse logic at `com.github.twitch4j.common.util.TwitchUtils#getPermissionsFromTags` for `global_mod` and `admin` badges?

iProdigy: `global_mod` is no longer used by twitch so you can skip that one. `admin` I've never seen in the wild so it's up to you. To make this change you'd also need to update `TwitchUtils.getPermissionsFromTags`

tadite:
Fixed problems from comments, deleted getters for global_mod and admin. Can you check, please?

iProdigy (approved):
Thank you again for your contribution

PhilippHeuer (approved):
Looks good now :)

PR #219, opened 2020-10-17
fix: register jackson modules despite exotic class loaders

iProdigy:

### Issues Fixed 
* Folks using Twitch4J in minecraft server plugins would have issues when `Instant`s needed to be deserialized from JSON, despite properly shading their plugin jar to include jackson and the jsr310 module. The custom [PluginClassLoader](https://papermc.io/javadocs/paper/1.13/org/bukkit/plugin/java/PluginClassLoader.html) used in this context likely is not compatible with `ObjectMapper#findAndRegisterModules`

#### Primary
* Explicitly register `ParameterNamesModule` and `JavaTimeModule` (so we are also shifting off of the deprecated `JSR310Module` with this PR) in `TypeConvert`, instead of relying upon `findAndRegisterModules`
* Use `TypeConvert`'s `ObjectMapper` wherever possible

#### Secondary
* Bump jackson version
* Bump caffeine version
* Bump JUnit version
* Fix typo in pagination of hype train events caused by Twitch's faulty docs
* Remove NonNull annotation on category search data due to [twitch inconsistency](https://github.com/twitchdev/issues/issues/238)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Plugins
 plugins {
-    id 'com.jfrog.bintray' version '1.8.5' apply false
-    id 'com.jfrog.artifactory' version '4.13.0' apply false
+	id 'com.jfrog.bintray' version '1.8.5' apply false
+	id 'com.jfrog.artifactory' version '4.13.0' apply false
 }
 
 // Artifact Info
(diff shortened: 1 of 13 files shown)


PhilippHeuer (approved):
Good work figuring this out :)

You missed one spot in the tmi module to set the new objectMapper, i added that one.

PR #220, opened 2020-10-18
feat: provide connection pooling helpers for pubsub and chat

iProdigy:

### Changes Proposed
* Create `TwitchPubSubConnectionPool`; allows for abstracted creation/deletion of `TwitchPubSub` instances in the background, depending on how many topics the user would like to subscribe to
* Create `TwitchChatConnectionPool`; allows for abstracted creation/deletion of `TwitchChat` instances in the background, depending on how many channels the user would like to join
* Create `TwitchChatLimitHelper` to store the standard bandwidth settings for various types of accounts
* Allow the same rate limit *buckets* (rather than simple bandwidth limits) to be shared across `TwitchChat` instances (important if they're using the same account)
* Add `autoJoinOwnChannel` as a configurable option for `TwitchChat`; useful for avoiding duplicated whisper events when using the chat pool
* Add `enableMembershipEvents` as a configurable option for `TwitchChat`; useful for reducing data transfer in constrained scenarios
* Bump lombok (necessary for SuperBuilder to function properly with generics)
* Bump events4j (necessary for multiple `TwitchChat` instances to work on a shared `EventManager`)


iProdigy:
Aside: I avoided applying `setterPrefix = "with"` to `@SuperBuilder`, because it is discouraged by the lombok maintainers, but I'm happy to deviate if you wish

common/src/main/java/com/github/twitch4j/common/pool/SubscriptionConnectionPool.java
```diff
@@ -0,0 +1,180 @@
...
+                unsaturatedConnections.putIfAbsent(c, 1);
+            else
+                saturatedConnections.add(c);
+        }
+        return c;
+    }
+
+    private void decrementSubscriptions(C connection) {
+        // Can no longer be saturated
+        saturatedConnections.remove(connection);
```
iProdigy, line 163: Because there is a short period of time after this removal takes place and `unsaturatedConnections` is not yet updated, it is possible for `getConnections`/`numConnections`/`numSubscriptions` to yield underestimates on concurrent threads as there is no synchronization. 

Scenarios where this could be problematic:
1. calling pool.close at the same time as a saturated connection is being decremented to unsaturated on another thread
2. tracking connection/subscription count in a multi-threaded test where enough `subscribe`s and `unsubscribe`s occur
3. implementing a pool with a max connections count restriction (though one would also need to synchronize the fallback code in `getOrCreateConnectionWithHeadroomAndIncrement` here)

Due to 1, it would be better to provide an overestimate if the lock-free approach were maintained. Here, dispose may just be called twice on the same connection, which is okay. 2 is not reflective of real-world usage and 3 already requires additional synchronization elsewhere

common/src/main/java/com/github/twitch4j/common/pool/TransactionalSubscriber.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Generic interface for objects that can respond to a subscription request, S,
+ * with a response, T, which can later be used for unsubscription to yield U.
+ *
+ * @param <S> subscription request
+ * @param <T> transactional response to the subscription request that can be used as an unsubscription request
+ * @param <U> unsubscription response
+ */
+public interface TransactionalSubscriber<S, T, U> {
```
iProdigy, line 11: Is this name appropriate? It may evoke the concept of database transactions that doesn't exactly apply here

pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSubConnectionPool.java
```diff
@@ -0,0 +1,79 @@
...
+public class TwitchPubSubConnectionPool extends TwitchModuleConnectionPool<TwitchPubSub, PubSubRequest, PubSubSubscription, Class<Void>, TwitchPubSubBuilder> {
+
+    private final String threadPrefix = "twitch4j-pool-" + RandomStringUtils.random(4, true, true) + "-pubsub-";
+
+    @Override
+    public PubSubSubscription subscribe(PubSubRequest pubSubRequest) {
+        final int topics = getTopicCount(pubSubRequest);
+        if (topics <= 0) return null;
+        if (topics > 1) throw new IllegalArgumentException("TwitchPubSubConnectionPool can only handle PubSubRequest's with a single topic subscription");
+        return super.subscribe(pubSubRequest);
```
iProdigy, line 28: If there was an error in subscription (e.g. BADAUTH), the connection ought to be decremented

pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSubConnectionPool.java
```diff
@@ -0,0 +1,79 @@
...
+            TwitchPubSubBuilder.builder()
+                .withEventManager(getConnectionEventManager())
+                .withScheduledThreadPoolExecutor(getExecutor(threadPrefix + RandomStringUtils.random(4, true, true), TwitchPubSub.REQUIRED_THREAD_COUNT))
+                .withProxyConfig(proxyConfig.get())
+        ).build();
+    }
+
+    @Override
+    protected void disposeConnection(TwitchPubSub connection) {
+        connection.disconnect();
```
iProdigy, line 43:
```suggestion
        connection.close();
```

chat/src/main/java/com/github/twitch4j/chat/TwitchChatConnectionPool.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Joins a channel.
+     *
+     * @param s the channel name
+     * @return the channel name
+     */
+    @Override
+    public String subscribe(String s) {
+        return super.subscribe(s != null ? s.toLowerCase() : null);
```
iProdigy, line 108: How should failures be handled (e.g. channel does not exist or user is banned from the chat)?

chat/src/main/java/com/github/twitch4j/chat/TwitchChatConnectionPool.java
```diff
@@ -0,0 +1,163 @@
...
+ * <ul>
+ *     <li>Reading from many channels without an account</li>
+ *     <li>Reading from many channels with an account</li>
+ *     <li>Reading from many channels using valid proxies</li>
+ *     <li>Reading and sending messages/whispers with only one account</li>
+ * </ul>
+ * Other applications are possible, but harder to configure. See below for a list of warnings.
+ * <p>
+ * Warning: If a custom {@link java.util.concurrent.ScheduledThreadPoolExecutor} is specified,
+ * its corePoolSize must be large enough for the threads required by connections made by this class.
```
iProdigy, line 25: Should make `TwitchChat` more thread-efficient (but this may be a task for another PR)

chat/src/main/java/com/github/twitch4j/chat/TwitchChatConnectionPool.java
iProdigy: We could immediately issue a leave or provide a boolean option to do nothing (could be useful as if the user is unbanned, for example, upon the next organic websocket reconnection, messages would be received again)

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -342,8 +324,8 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
         log.debug("Registering the following command triggers: " + commandPrefixes.toString());
 
         // register event handler
-        eventManager.onEvent(ChannelMessageEvent.class, this::onChannelMessage);
-        eventManager.onEvent(IRCMessageEvent.class, event -> {
+        eventManager.onEvent("twitch4j-chat-command-trigger", ChannelMessageEvent.class, this::onChannelMessage);
+        eventManager.onEvent("twitch4j-chat-id-name-cache", IRCMessageEvent.class, event -> {
```
PhilippHeuer, line 328: This would only register the id-name cache function once, so it would only work for a single twitchChat instance and not on the other connections right?

iProdigy: Yeah for each EventManager (that is associated with at least one TwitchChat instance), we only want these event consumers to be registered once or else the consumption would be duplicated (and this could result in firing two CommandEvents when only one is appropriate, for example). With the connection pool, multiple TwitchChat instances are sharing the same EventManager, so these unique ids are necessary. Let me know if I missed the point of your question

PhilippHeuer:
What do you think about creating a shared interface between TwitchChat / TwitchChatConnectionPool and TwitchPubSub / TwitchPubSubConnection Pool?

This would allow us to add connection pool configuration parameters to the clientBuilder (enable, number of `subscriptions` per connection for chat / pubsub) and if enabled we could return a connection pool instead, and by returning the interface in the builder the code would work with and without connection pool.

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
PhilippHeuer: This was only supposed to highlight the 2nd line.

`twitch4j-chat-id-name-cache` registers a consumer function in one connection (twitchChat instance) that writes the cache entries into a map in that specific connection - (`channelIdToChannelName` and `channelNameToChannelId `), so those maps would be empty in all other connections, except for the first one as the consumer function is only registered once.

Those maps are used in each connection individually when firing the IRCMessageEvent, so in all connections except for the first one the code that uses those map in the IRCMessageEvent wouldn't work as that will always get a empty map.

iProdigy: Ahh, thanks for the clarification - that's a good point. Unfortunately, removing the uniqueness would result in each connection also caching channel id/name info for other TwitchChat instances on the same EventManager. A naive sharing of the map across instances would run into synchronization issues... Should we switch to the less-performance synchronizedMap to be shared across instances or not worry about cache duplication

iProdigy: Perhaps each TwitchChat instance could have a dynamically generated UUID that is included in the IRCMessageEvent. Then, when it comes time to cache id/name, the consumer would first check if the UUID of the event and the UUID of the instance match

PhilippHeuer: I'm not worried about cache duplication in this case, the map just holds channel id & name, even with a lot of channels it will have a very small footprint - therefore i would just remove the uniqueness for the consumer.

iProdigy:
> This would allow us to add connection pool configuration parameters to the clientBuilder (enable, number of `subscriptions` per connection for chat / pubsub) and if enabled we could return a connection pool instead, and by returning the interface in the builder the code would work with and without connection pool.

I like the idea of tying the connection pools to the existing builders rather than existing as a totally independent concept (but if they want to use more complicated configurations, they'll have to use the specialized builders). 

Maybe:
```java
TwitchClientBuilder.builder()
  .withEnableChatPool(true) // implicitly also does withEnableChat(true)
  .withEnablePubSubPool(true) // implicitly also does withEnablePubSub(true)
  .withMaxSubscriptionsPerPoolConnection(50) // applies to both chat & pubsub? or we can have two
  .build();
```

Then, in `TwitchClient`, we offer direct access to the individual pool objects. Also, `getChat()` and `getPubSub` can return dummy objects that just delegate to the pool.

PhilippHeuer:
> ```java
> TwitchClientBuilder.builder()
>   .withEnableChatPool(true) // implicitly also does withEnableChat(true)
>   .withEnablePubSubPool(true) // implicitly also does withEnablePubSub(true)
>   .withMaxSubscriptionsPerPoolConnection(50) // applies to both chat & pubsub? or we can have two
>   .build();
> ```

Yeah thats a good idea, the clientBuilder should just have the toggle and size, for anything more complex one has to use the pool builder. I would split it up into chat / pubsub size.

> Then, in `TwitchClient`, we offer direct access to the individual pool objects. Also, `getChat()` and `getPubSub` can return dummy objects that just delegate to the pool... but this needs a shared interface to implement & changing the return type of getChat to return the interface would be a breaking change?

Agree, changing the return type now would be a breaking change - if only i created interfaces for each module right at the start :(

How about adding a `TwitchClientPoolBuilder` that has the additional properties and returns the new interface instead?
We could then change the return type in the ClientBuilder to the interface when we publish the next major release.

twitch4j/src/main/java/com/github/twitch4j/modules/IModule.java
```diff
@@ -10,7 +10,7 @@
 	private final String name;
 	private final String author;
 	private final String version;
-	public abstract void enable(TwitchClient client);
+	public abstract void enable(ITwitchClient client);
```
iProdigy, line 13: Is this fine?

twitch4j/src/main/java/com/github/twitch4j/TwitchClientPoolBuilder.java
```diff
@@ -0,0 +1,447 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+
+/**
+ * Builder to get a TwitchClientPool Instance by provided various options, to provide the user with a lot of customizable options.
+ */
+@Slf4j
+@Getter
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class TwitchClientPoolBuilder {
```
iProdigy, line 57: Should this share a common interface or abstract class with TwitchClientBuilder?

twitch4j/src/main/java/com/github/twitch4j/modules/IModule.java
PhilippHeuer: Yes thats fine

twitch4j/src/main/java/com/github/twitch4j/TwitchClientPoolBuilder.java
PhilippHeuer: We don't need that here, that class is only a workaround because we don't want to change the return types in the TwitchClientBuilder, so this class will be removed with the next major release and those 4 new properties will go into the TwitchClientBuilder.

PhilippHeuer (approved):
LGTM

PR #223, opened 2020-10-25
build: gradle build updates

stachu540:

### Changes Proposed

- update gradle to 6.7
- create sourcesJar and javadocJar using JavaPluginExtension - [reference](https://docs.gradle.org/current/samples/sample_building_java_libraries.html)
- lower-case archive base name

### Additional Information 
this commit provides gradle updates to 6.7 and using JavaPluginExtension to creating sourcesJar and javadocJar. Additonal change is usage `tasks.withType(Jar)` to define `archiveBaseName`. `baseName` is deprecated for removal since 7.0.

Next stop is combined javadoc with delombok.


PhilippHeuer (changes requested):
It's a good idea to use the default sources / javadoc to be conform with the latest gradle version.

Can you remove this block: https://github.com/twitch4j/twitch4j/blob/5129608483db9da4c8172bb9db7f68a626cdd1ee/deployment.gradle#L55-L61

and those 2 lines: https://github.com/twitch4j/twitch4j/blob/5129608483db9da4c8172bb9db7f68a626cdd1ee/deployment.gradle#L91-L92

Those would break the deployment process, with the new tasks we no longer have to specify the additional artifacts for sources / javadocs.

stachu540:
Done

PhilippHeuer (approved):

PR #229, opened 2020-11-10
build(ci): migration to github actions

stachu540:

### Changes Proposed

* Migration CI/CD to GitHub Actions

### Additional Information
This migration contains:
 - `ci` - which are testing lib in JDK's: 8, 11, 14 and 15 when someone push to some branches or make a PR
 - `release` - works only if you create tag in `master` branch. with [Release Drafter](https://github.com/release-drafter/release-drafter) will create release contains with specific label will be sorted and ordered. It required define `BINTRAY_USER` and `BINTRAY_API_KEY` secrets in repository settings.
 - `publication` - will notify to the discord channel via webhooks. Required define `DISCORD_WEBHOOK` secrets in repository settings.

`GITHUB_TOKEN` secret contains token using by [@github-actions](https://github.com/apps/github-actions) bot

Also there is:
- `release-drafter.yml` - for build Releases. See: [Release Drafter](https://github.com/release-drafter/release-drafter)
- `pr-labeler.yml` - helps automatically label PR's when have some specific conditions. Requires [actions/labeler](https://github.com/actions/labeler) to use it. But there is other way solutions to use [Community Health File](https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/creating-a-default-community-health-file#about-default-community-health-files) directed to `.github` repository for each organizations or users. It will assign automatically without using GH Actions defined inside Front Matter template.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: CI
+on: [ push, pull_request ]
+
+jobs:
+  ci:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jvm: [ 8, 11, 14, 15 ]
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout
+        with:
+          fetch-depth: 0    # Fetch all history for .GitInfo and .Lastmod
+      - name: Prepare cache builds
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: JDK${{ matrix.jvm }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: JDK${{ matrix.jvm }}-gradle-
+      - uses: actions/setup-java@v1
+        name: Setup JDK ${{ matrix.jvm }}
+        with:
+          java-version: ${{ matrix.jvm }}
+      - name: Prepare to Test
+        run: chmod +x ./gradlew
+      - name: Test Unit
+        run: ./gradlew test
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,3 @@
 /.github export-ignore
 /.gitattributes export-ignore
 /.editorconfig export-ignore
-/circle.yml export-ignore
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,0 +1,3 @@
+feature: ['feature/*', 'feat/*']
+bugfix: fix/*
+deprecated: deprecate/*
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,30 @@
+name-template: Twitch4J v$RESOLVED_VERSION
+categories:
+  - title: New Features
+    labels:
+      - feature
+      - enhancement
+  - title: Bugs Fixed
+    labels:
+      - 'bug/minor'
+      - 'bug/major'
+      - bugfix
+  - title: Deprecations
+    label: deprecated
+exclude-labels:
+  - wontfix
+replacers:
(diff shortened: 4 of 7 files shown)


stachu540:
I can use semantic version using this: https://github.com/marketplace/actions/action-for-semantic-release
rather than [release drafter](https://github.com/marketplace/actions/release-drafter). Action uses [Semantic Release](https://github.com/semantic-release/semantic-release) as their plugins in [this org](https://github.com/semantic-release).

stachu540:
GH actions has been ready to review. Some automation will be theoretically worked.

There is more feature has been updated in the description. Feel free to check it out.

stachu540:
I have update one more time. Now will checks approved reviews. If reviews will be more then 1 it will auto-merge using title as commit message (of course will squash commit).

stachu540:
I will temporary stale this PR before testing it myself. After that i will give feedback and pulling some changes in this branch.

stachu540:
The changes has been applied. Don't know how about the pull request's action. Anything else leaving to review for @PhilippHeuer

gradle.properties
```diff
@@ -0,0 +1,2 @@
+group = com.github.twitch4j
+version = 1.2.1
```
iProdigy, line 2:
```suggestion
version = 1.3.2
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+    branches: [ develop ]
+    paths-ignore:
+      - ".github/**"
+      - "**.md"
+      - ".gitignore"
+      - ".editorconfig"
+
+env:
+  RELEASE: false
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
PhilippHeuer, line 20: I don't think the GH TOKEN is needed here - also we can't use any secrets in workflows that run for pull requests.

gradle.properties
PhilippHeuer: We don't need to update this manually, the version property should always be set as parameter from ci. A placeholder like 0.0.0 is fine, as the default version in here will never be used.

stachu540: no nned t attaching version. as @PhilippHeuer said 0.0.0 i enough for that

.github/workflows/ci.yml
stachu540: This should be not needed in test unit. But if you push something into gh pages, registry or the release

iProdigy (changes requested):
Should explicitly declare `artifactId` (see: https://github.com/twitch4j/twitch4j/pull/243#issuecomment-829152764)

Even this change would not be sufficient (e.g. the tmi module would be named `twitch4j-tmi`, which is better than `twitch4j-rest-tmi`, but not quite the desired `twitch4j-messaginginterface`):
```kotlin
archivesBaseName = artifactId.let {
	val prefixCheck = rootProject.artifactId + "-rest-"
	if (it.startsWith(prefixCheck, true))
		rootProject.artifactId + "-" + it.substring(prefixCheck.length)
	else
		it
}
```

stachu540:
removing `rest-` in module namings will resolves your request

PhilippHeuer:
I do prefer to group all the api modules with the rest prefix, but we can't change the artifactId's for already released modules - so that one line in each module to set the artifactId is fine.

iProdigy (dismissed):

PhilippHeuer:
for the gradle-wrapper workflow:

I added `base-branch` because it ended up pulling from master/main anyway - it's better to work fully on develop. And i also added the option to manually start the workflow. Example PR here: https://github.com/twitch4j/twitch4j-ghactions-test/pull/28

Next to test will be the MR workflows.

stachu540:
Unfortunate, this action will not have a pull request title definition to when to be a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/). Wish we could do something of that.

PR #230, opened 2020-11-30
Fix userId being undefined

C0D3-M4513R:
When creating an OAuth Token for twitch, using CredentialManager, TwitchIdentityProdider and a user-id/user-secret,  getAdditionalCredentialInformation would return a OAuth2Credential, that didn't have a userId set. The effects are, that in the CredentialManager, the getOAuth2CredentialByUserId Method doesn't work. This Pull Request Changes this, by actually getting the right key in the request.

* [x] I have tested this feature - See https://github.com/twitch4j/twitch4j/blob/master/auth/src/main/java/com/github/twitch4j/auth/providers/TwitchIdentityProvider.java#L69

...

### Changes Proposed

* getAdditionalCredentialInformation is now returning an actual userId, rather than null

...


iProdigy:
A user id and client id are separate concepts; Client IDs identify applications while User IDs identify individual users. With tokens, they can either be *user* access tokens or *app* access tokens. In the latter case, there is no user attached, so `user_id` would be null — this is the situation if you are obtaining your token through `TwitchIdentityProvider#getAppAccessToken`.

PR #232, opened 2020-12-19
feat: add channel point predictions pubsub and badge

iProdigy:

### Changes Proposed
* Implement `predictions-channel-v1` unofficial pubsub topic
* Implement `predictions-user-v1` unofficial pubsub topic
* Implement `CommandPermission.PREDICTIONS_BLUE` and `CommandPermission.PREDICTIONS_PINK`

### Additional Information 
https://blog.twitch.tv/en/2020/12/12/channel-points-predictions-let-your-viewers-guess-your-destiny/

### DIFF
--- a/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/Prediction.java
+++ b/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/Prediction.java
@@ -1,0 +1,27 @@
+package com.github.twitch4j.pubsub.domain;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.Setter;
+
+import java.time.Instant;
+
+@Data
+@Setter(AccessLevel.NONE)
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Prediction {
+    private String id;
+    private String eventId;
+    private String outcomeId;
+    private String channelId;
+    private Integer points;
+    private Instant predictedAt;
+    private Instant updatedAt;
+    private String userId;
+    private PredictionResult result; // can be null
+    private String userDisplayName;
+}
--- a/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/Prediction.java
+++ b/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/Prediction.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import lombok.AccessLevel;
 import lombok.Data;
 import lombok.Setter;
+import org.jetbrains.annotations.Nullable;
 
 import java.time.Instant;
 
@@ -22,6 +23,7 @@ public class Prediction {
     private Instant predictedAt;
     private Instant updatedAt;
     private String userId;
-    private PredictionResult result; // can be null
+    @Nullable
+    private PredictionResult result;
     private String userDisplayName;
 }
--- a/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionEvent.java
+++ b/pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionEvent.java
@@ -1,0 +1,31 @@
+package com.github.twitch4j.pubsub.domain;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.Setter;
+
+import java.time.Instant;
+import java.util.List;
+
+@Data
+@Setter(AccessLevel.NONE)
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class PredictionEvent {
+    private String id;
+    private String channelId;
+    private Instant createdAt;
+    private PredictionTrigger createdBy;
+    private Instant endedAt;
+    private PredictionTrigger endedBy;
+    private Instant lockedAt;
+    private PredictionTrigger lockedBy;
+    private List<PredictionOutcome> outcomes;
+    private Integer predictionWindowSeconds;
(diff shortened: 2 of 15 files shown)


pubsub/src/main/java/com/github/twitch4j/pubsub/domain/Prediction.java
```diff
@@ -0,0 +1,27 @@
...
+public class Prediction {
+    private String id;
+    private String eventId;
+    private String outcomeId;
+    private String channelId;
+    private Integer points;
+    private Instant predictedAt;
+    private Instant updatedAt;
+    private String userId;
+    private PredictionResult result; // can be null
```
PhilippHeuer, line 25: Can we annotate this with @ Nullable instead of the comment? Lombok should pass that annotation to the getter/setter method.
> lombok.copyableAnnotations = [A list of fully qualified types] (default: empty list)
Lombok will copy any of these annotations from the field to the setter parameter, and to the getter method. Note that lombok ships with a bunch of annotations 'out of the box' which are known to be copyable: All popular nullable/nonnull annotations.

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionEvent.java
```diff
@@ -0,0 +1,31 @@
...
+    private String channelId;
+    private Instant createdAt;
+    private PredictionTrigger createdBy;
+    private Instant endedAt;
+    private PredictionTrigger endedBy;
+    private Instant lockedAt;
+    private PredictionTrigger lockedBy;
+    private List<PredictionOutcome> outcomes;
+    private Integer predictionWindowSeconds;
+    private String status; // e.g. ACTIVE or LOCKED or RESOLVE_PENDING or RESOLVED
```
PhilippHeuer, line 28: Can we use a enum here as we only have 4 possible values?

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionResult.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+@Data
+@Setter(AccessLevel.NONE)
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class PredictionResult {
+
+    private String type; // e.g. WIN or LOSE
```
PhilippHeuer, line 18: Can we use a enum for this too (WIN/LOSE)?

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionResult.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private Integer pointsWon; // can be null
```
PhilippHeuer, line 20: Lets use the nullable annotation here too

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionEvent.java
iProdigy: I think there is at least one more state (e.g. if the prediction was cancelled before being resolved) - I can figure out what `status` is in this scenario but I was afraid of more possible states as this topic is undocumented

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionResult.java
iProdigy: I believe there could be at least one more state for this variable as well; if a prediction isn't resolved within 24 hours, the points are refunded, so `type` may be something special in this case

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionResult.java
stachu540: I'd recommend use [JRS-305](https://mvnrepository.com/artifact/com.google.code.findbugs/jsr305/3.0.2) for all repo. Will be better markable as nullable.

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionEvent.java
PhilippHeuer: Then lets keep it as comment on the field as it is currently 👍

iProdigy: Should we put this info in a javadoc comment to make it more visible or not bother due to the undocumented nature?

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionResult.java
PhilippHeuer: Alright, on 2nd thought maybe using strings for states on unofficial topics is a better choice in general in case they change states unannaunced - so this would be more resilient to changes.

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionEvent.java
PhilippHeuer: Good idea, lets put all currently known states into the javadoc comment - maybe someone discovers the states there before asking what values are possible in discord

stachu540: > javadoc on the field will now be copied to generated getters and setters. Normally, all text is copied, and `@return` is moved to the getter, whilst `@param` lines are moved to the setter. Moved means: Deleted from the field's javadoc. It is also possible to define unique text for each getter/setter. To do that, you create a 'section' named `GETTER` and/or `SETTER`. A section is a line in your javadoc containing 2 or more dashes, then the text 'GETTER' or 'SETTER', followed by 2 or more dashes, and nothing else on the line. If you use sections, `@return` and `@param` stripping for that section is no longer done (move the `@return` or `@param` line into the section).

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/PredictionResult.java
PhilippHeuer: We haven't really used the nullable/notnull annotation before so we're free to pick i think, the auto generated code in graphql uses the jetbrains annotation dependency.

Is there a reason why we should use JSR305 over the jetbrains annotation dependency?

stachu540: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003485599/comments/360000474759 - up to you

PhilippHeuer (approved):

PR #302, opened 2021-03-20
feat: add unofficial badge endpoints

iProdigy:

### Changes Proposed
* Add `TwitchKraken#getGlobalBadges(String)`
* Add `TwitchKraken#getChannelBadges(String, String)`

### Additional Information 
https://discuss.dev.twitch.tv/t/beta-badge-api/6388

### DIFF
--- a/rest-kraken/src/main/java/com/github/twitch4j/kraken/TwitchKraken.java
+++ b/rest-kraken/src/main/java/com/github/twitch4j/kraken/TwitchKraken.java
@@ -34,56 +34,6 @@ public interface TwitchKraken {
     }).get();
 
     /**
-     * The default baseUrl to pass to {@link #getGlobalBadges(URI, String)} and {@link #getChannelBadges(URI, String, String)}
-     */
-    @Unofficial
-    URI BADGES_BASE_URL = ((Supplier<URI>) () -> {
-        try {
-            return new URI("https://badges.twitch.tv/v1");
-        } catch (Exception e) {
-            return null;
-        }
-    }).get();
-
-    @Unofficial
-    @RequestLine("GET /badges/global/display?language={language}")
-    HystrixCommand<BadgeSets> getGlobalBadges(
-        URI baseUrl,
-        @Param("language") String language
-    );
-
-    /**
-     * Get the chat badges that are globally available in chat.
-     *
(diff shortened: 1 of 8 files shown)


rest-kraken/src/main/java/com/github/twitch4j/kraken/TwitchKraken.java
```diff
@@ -33,6 +33,56 @@
...
+        try {
+            return new URI("https://badges.twitch.tv/v1");
+        } catch (Exception e) {
+            return null;
+        }
+    }).get();
+
+    @Unofficial
+    @RequestLine("GET /badges/global/display?language={language}")
+    HystrixCommand<BadgeSets> getGlobalBadges(
```
PhilippHeuer, line 50: Considering we're on java 8 i don't think we can make the helper methods private/not expose them - maybe we can deprecate them to discourage users from trying to use them?

rest-kraken/src/main/java/com/github/twitch4j/kraken/TwitchKraken.java
```diff
@@ -33,6 +33,56 @@
...
+     */
+    @Unofficial
+    default HystrixCommand<BadgeSets> getGlobalBadges(String language) {
+        return getGlobalBadges(BADGES_BASE_URL, language);
+    }
+
+    @Unofficial
+    @RequestLine("GET /badges/channels/{channel_id}/display?language={language}")
+    HystrixCommand<BadgeSets> getChannelBadges(
+        URI baseUrl,
```
PhilippHeuer, line 69: As above, maybe deprecate this method too.

PhilippHeuer (changes requested):
I think the badges code might be a better fit for the tmi module as it's related to chat and isn't a kraken method. Maybe we could use tmi for any future api methods that are not kraken or helix.

iProdigy:
> I think the badges code might be a better fit for the tmi module as it's related to chat and isn't a kraken method. Maybe we could use tmi for any future api methods that are not kraken or helix.

Agreed; especially as twitch seems to be killing some of the current tmi endpoints so the module may become otherwise useless

PhilippHeuer (approved):

PR #306, opened 2021-04-07
feat: automatically handle helix rate limits

iProdigy:

### Changes Proposed
* Track api bucket point usage of the user/app access token(s) in `TwitchHelixClientIdInterceptor`
* Create a custom `JacksonDecoder` to track ratelimit headers received by twitch to ensure our bucket cache is in harmony
* Delegate http calls in `TwitchHelixHttpClient` immediately or with delay, depending on the amount of points left in the corresponding bucket

### DIFF
--- a/rest-helix/build.gradle
+++ b/rest-helix/build.gradle
@@ -7,6 +7,9 @@ dependencies {
 	api group: 'io.github.openfeign', name: 'feign-hystrix'
 	api group: 'commons-configuration', name: 'commons-configuration'
 
+	// Rate Limiting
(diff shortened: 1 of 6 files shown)


iProdigy:
The main downside of this approach: https://github.com/twitch4j/twitch4j/blob/feature/helix-rate-limiting/rest-helix/src/main/java/com/github/twitch4j/helix/interceptor/TwitchHelixHttpClient.java#L51

* `withTimeout` needs to be high enough for the call not to fail while waiting for bucket headroom
* the waiting is blocking rather than via elegant executor scheduling

Unfortunately, I'm not sure if feign/hystrix allows a better way

PhilippHeuer (approved):

PR #332, opened 2021-05-08
feat: handle official user-moderation-notifications pubsub topic

iProdigy:

### Changes Proposed
* Implement official `user-moderation-notifications.<user-id>.<channel-id>`  pubsub topic

### Additional Information
Depends on #331

### DIFF
--- a/pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSub.java
+++ b/pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSub.java
@@ -557,6 +557,16 @@ public class TwitchPubSub implements ITwitchPubSub {
                                         log.warn("Unparsable Message: " + message.getType() + "|" + message.getData());
                                         break;
                                 }
+                            } else if (topic.startsWith("user-moderation-notifications")) {
+                                int lastPeriod = topic.lastIndexOf('.');
+                                String userId = topic.substring(topic.indexOf('.') + 1, lastPeriod);
+                                String channelId = topic.substring(lastPeriod + 1);
+                                if ("automod_caught_message".equalsIgnoreCase(type)) {
+                                    UserAutomodCaughtMessage data = TypeConvert.convertValue(msgData, UserAutomodCaughtMessage.class);
+                                    eventManager.publish(new UserAutomodCaughtMessageEvent(userId, channelId, data));
+                                } else {
+                                    log.warn("Unparsable Message: " + message.getType() + "|" + message.getData());
+                                }
                             } else if (topic.startsWith("polls")) {
                                 PollData pollData = TypeConvert.convertValue(msgData.path("poll"), PollData.class);
                                 eventManager.publish(new PollsEvent(type, pollData));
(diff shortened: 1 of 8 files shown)


pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSub.java
```diff
@@ -564,6 +564,16 @@ public void onTextMessage(WebSocket ws, String text) {
                                         log.warn("Unparsable Message: " + message.getType() + "|" + message.getData());
                                         break;
                                 }
+                            } else if (topic.startsWith("user-moderation-notifications")) {
+                                int lastPeriod = topic.lastIndexOf('.');
+                                String userId = topic.substring(topic.indexOf('.') + 1, lastPeriod);
```
PhilippHeuer, line 569: Maybe we can improve this a little by adding sth. after https://github.com/twitch4j/twitch4j/blob/39df41f85ce6cdc6810c4cdeb04a347aa0bbb1e4/pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSub.java#L369 to parse the topic value first (split on `.`).

And then check for a exact match on the topic name and check that the param count (ie. 3 for this one, topic + userId + channelId) matches.
We wouldn't need the lastIndexOf / substring code in every case then.

PhilippHeuer (changes requested):

pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSub.java
iProdigy: Good idea, I switched to the split approach for this topic, will migrate the others in a separate refactor PR

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/AutomodContentClassification.java
```diff
@@ -8,11 +9,50 @@
...
+        // New Categories
+        AGGRESSION,
+        BULLYING,
+        DISABILITY,
+        MISOGYNY,
+        RACE_ETHNICITY_OR_RELIGION,
+        SEXUALITY_SEX_OR_GENDER,
+        SEX_BASED_TERMS,
+        SWEARING,
+        // Possible Categories
```
iProdigy, line 32: Could do some manual parsing work here to consolidate enum values

pubsub/src/main/java/com/github/twitch4j/pubsub/domain/AutomodCaughtMessage.java
```diff
@@ -35,15 +63,37 @@
     @Setter(AccessLevel.PRIVATE)
     public static class FragmentFlags {
         private Map<String, Integer> topics;
+
+        public Map<AutomodContentClassification.Category, Integer> getParsedTopics() {
+            return TypeConvert.convertValue(topics, new TypeReference<EnumMap<AutomodContentClassification.Category, Integer>>() {});
```
iProdigy, line 68: technically not the most performant solution, but this supports `@JsonAlias`

PhilippHeuer (approved):

PR #386, opened 2021-06-22
feat: implement new helix schedule api

iProdigy:

### Changes Proposed
* Add `TwitchHelix#getChannelStreamSchedule`
* Add `TwitchHelix#getChannelInternetCalendar`
* Add `TwitchHelix#updateChannelStreamSchedule`
* Add `TwitchHelix#createStreamScheduleSegment`
* Add `TwitchHelix#updateStreamScheduleSegment`
* Add `TwitchHelix#deleteStreamScheduleSegment`

### Additional Information
https://dev.twitch.tv/docs/change-log

### DIFF
--- a/rest-helix/src/main/java/com/github/twitch4j/helix/TwitchHelix.java
+++ b/rest-helix/src/main/java/com/github/twitch4j/helix/TwitchHelix.java
@@ -991,6 +991,113 @@ public interface TwitchHelix {
     );
 
     /**
+     * Gets all scheduled broadcasts or specific scheduled broadcasts from a channel’s stream schedule.
+     *
+     * @param authToken     User OAuth Token or App Access Token.
+     * @param broadcasterId User ID of the broadcaster who owns the channel streaming schedule.
+     * @param ids           The ID of the stream segment to return.Maximum: 100.
+     * @param startTime     A timestamp in RFC3339 format to start returning stream segments from. If not specified, the current date and time is used.
+     * @param utcOffset     A timezone offset for the requester specified in minutes. For example, a timezone that is +4 hours from GMT would be “240.” If not specified, “0” is used for GMT.
+     * @param after         Cursor for forward pagination: tells the server where to start fetching the next set of results in a multi-page response. The cursor value specified here is from the pagination response field of a prior query.
+     * @param limit         Maximum number of stream segments to return.Maximum: 25. Default: 20.
+     * @return StreamScheduleResponse
+     */
+    @RequestLine("GET /schedule?broadcaster_id={broadcaster_id}&id={id}&start_time={start_time}&utc_offset={utc_offset}&first={first}&after={after}")
+    @Headers("Authorization: Bearer {token}")
+    HystrixCommand<StreamScheduleResponse> getChannelStreamSchedule(
+        @Param("token") String authToken,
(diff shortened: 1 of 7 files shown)


rest-helix/src/main/java/com/github/twitch4j/helix/TwitchHelix.java
```diff
@@ -990,6 +990,113 @@
...
+    );
+
+    /**
+     * Gets all scheduled broadcasts from a channel’s stream schedule as an iCalendar.
+     *
+     * @param broadcasterId User ID of the broadcaster who owns the channel streaming schedule.
+     * @return iCalendar data is returned according to RFC5545. The expected MIME type is text/calendar.
+     */
+    @RequestLine("GET /schedule/icalendar?broadcaster_id={broadcaster_id}")
+    HystrixCommand<Response> getChannelInternetCalendar(
```
iProdigy, line 1024: Is there a better return type possible here?

rest-helix/src/main/java/com/github/twitch4j/helix/TwitchHelix.java
```diff
@@ -990,6 +990,113 @@
...
+    /**
+     * Gets all scheduled broadcasts or specific scheduled broadcasts from a channel’s stream schedule.
+     *
+     * @param authToken     User OAuth Token or App Access Token.
+     * @param broadcasterId User ID of the broadcaster who owns the channel streaming schedule.
+     * @param ids           The ID of the stream segment to return.Maximum: 100.
+     * @param startTime     A timestamp in RFC3339 format to start returning stream segments from. If not specified, the current date and time is used.
+     * @param utcOffset     A timezone offset for the requester specified in minutes. For example, a timezone that is +4 hours from GMT would be “240.” If not specified, “0” is used for GMT.
+     * @param after         Cursor for forward pagination: tells the server where to start fetching the next set of results in a multi-page response. The cursor value specified here is from the pagination response field of a prior query.
+     * @param limit         Maximum number of stream segments to return.Maximum: 25. Default: 20.
```
iProdigy, line 1002:
```suggestion
     * @param ids           The ID of the stream segment to return. Maximum: 100.
     * @param startTime     A timestamp in RFC3339 format to start returning stream segments from. If not specified, the current date and time is used.
     * @param utcOffset     A timezone offset for the requester specified in minutes. For example, a timezone that is +4 hours from GMT would be “240.” If not specified, “0” is used for GMT.
     * @param after         Cursor for forward pagination: tells the server where to start fetching the next set of results in a multi-page response. The cursor value specified here is from the pagination response field of a prior query.
     * @param limit         Maximum number of stream segments to return. Maximum: 25. Default: 20.
```

rest-helix/src/main/java/com/github/twitch4j/helix/TwitchHelix.java
stachu540: Possible usage with: https://github.com/ical4j/ical4j

iProdigy: Yeah but I also don't want to add a dependency for just that endpoint - though, I probably should add a default method that converts the Response to a String

PhilippHeuer (approved):

PR #418, opened 2021-08-06
refactor: allow chat limit specification in more builders

iProdigy:

### Changes Proposed
* Create `TwitchLimitRegistry` for people to specify elevated rate-limit bandwidths for specific user id's
* Ensure chat message/join/whisper default limit can be configured in each relevant builder
* In TwitchChatBuilder, obtain the shared rate-limit buckets from `TwitchLimitRegistry`

### Additional Information
Recently twitch actually started enforcing it's irc JOIN ratelimit (rather strictly). As a result, there has been more demand to be able to easily override the default non-message ratelimits (e.g., to set a lower limit than the default to avoid network oddities making it appear that a small ratelimit violation occurred from twitch's perspective). Also, the bot verification process has started again so more folks may want to easily specify their updated ratelimits.

### DIFF
--- a/common/src/main/java/com/github/twitch4j/common/util/TwitchLimitRegistry.java
+++ b/common/src/main/java/com/github/twitch4j/common/util/TwitchLimitRegistry.java
@@ -5,7 +5,9 @@ import io.github.bucket4j.Bandwidth;
 import io.github.bucket4j.Bucket;
 import io.github.bucket4j.Bucket4j;
 import io.github.bucket4j.BucketConfiguration;
+import io.github.bucket4j.IncompatibleConfigurationException;
 import io.github.bucket4j.local.LocalBucketBuilder;
+import lombok.NonNull;
 
 import java.util.Collections;
 import java.util.EnumMap;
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/github/twitch4j/common/util/TwitchLimitRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+    private static Bucket constructBucket(List<Bandwidth> limits) {
+        LocalBucketBuilder builder = Bucket4j.builder();
+        limits.forEach(builder::addLimit);
+        return builder.build();
+    }
+
+    /**
+     * @return the single thread-safe instance of the limit registry.
+     */
+    public static TwitchLimitRegistry getInstance() {
```
iProdigy, line 97: Is the singleton approach fine or would you prefer a static helper class?

common/src/main/java/com/github/twitch4j/common/util/TwitchLimitRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Attempts to set a user's bucket for a specific rate-limit.
+     *
+     * @param userId    the id of the user whose rate limit is being specified.
+     * @param limitType the type of rate-limit that is being configured.
+     * @param limit     the bandwidths that are applicable for this user and rate limit type.
+     * @throws IncompatibleConfigurationException if overriding an old limit that is incompatible with the new configuration
+     */
+    public void setLimit(@NonNull String userId, @NonNull TwitchLimitType limitType, @NonNull List<Bandwidth> limit) {
```
iProdigy, line 49: Should there be versions where `List<Bandwidth>` is replaced with `Bandwidth[]` or `Bandwidth` (or is it fine to make the user call `Arrays.asList` or `Collections.singletonList` respectively)

common/src/main/java/com/github/twitch4j/common/util/TwitchLimitRegistry.java
PhilippHeuer: This approach is fine.

common/src/main/java/com/github/twitch4j/common/util/TwitchLimitRegistry.java
PhilippHeuer: I think it's fine, there is no need to overload this method.

PhilippHeuer (approved):

PR #440, opened 2021-10-01
chore(jsonProps): edit value of json properties to match lower camel case

guillaume-houioux:

### Changes Proposed

* change value of json properties that are not in lower camel case in `Stream.java`  class


iProdigy:
The variables themselves are already in lowerCamelCase (`startedAtInstant` & `isMature`). The `@JsonProperty` annotation is saying that the property names in json format are not in the same form. In particular, it comes from the reference docs for that endpoint: https://dev.twitch.tv/docs/api/reference#get-streams

PR #487, opened 2021-12-19
chore: apply ident / import policies to all files

PhilippHeuer:

### Changes Proposed

* a lot of files were still using tab-ident instead of the whitespace defined in .editorconfig
* always use explicit imports instead of * imports
* optimize imports in all files
* require whitespace in array init

#### Additional notes
This PR only contains formatting changes and no code changes.

Checkstyle: https://github.com/PhilippHeuer/gradle-projectconfiguration/blob/main/plugin/src/main/resources/checkstyle/twitch4j.xml


common/src/main/java/com/github/twitch4j/common/util/EscapeUtils.java
```diff
@@ -70,14 +70,14 @@ public static String escapeTagValue(Object value) {
     public static String unescapeTagValue(String value) {
         return StringUtils.replaceEach(
             value,
-            new String[] {
+            new String[]{
```
iProdigy, line 73: I personally [prefer](https://google.github.io/styleguide/javaguide.html#s4.8.3.1-array-initializers) the space but we can remove it

eventsub-common/src/main/java/com/github/twitch4j/eventsub/EventSubSubscription.java
```diff
@@ -43,7 +43,7 @@
      */
     @JsonIgnore
     @ToString.Exclude
-    @Getter(onMethod_ = { @JsonIgnore })
+    @Getter(onMethod_ = {@JsonIgnore})
```
iProdigy, line 46: though this space removal i'm cool with 👍

PR #503, opened 2022-01-07
feat: retry failed joins in irc with exponential backoff

iProdigy:

### Issues Fixed 
* Should help with phantom joined channels

### Changes Proposed
* On initial join attempt, put channel name in `joinAttemptsByChannelName`
* On successful join, remove channel name from `joinAttemptsByChannelName`
* If entry expires from `joinAttemptsByChannelName`, the join failed. Try again with exponential backoff (if maximum retries has not been reached)

* Add `removeChannelOnJoinFailure` option to `TwitchChat(Builder)`. Upon such a removal, fire `ChannelRemovedPostJoinFailureEvent`
* Add `maxJoinRetries` to builders

### Additional Information
https://github.com/ben-manes/caffeine/wiki/Removal

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
@@ -409,7 +409,7 @@ public class TwitchChat implements ITwitchChat {
                         } else if (removeChannelOnJoinFailure && removeCurrentChannel(name)) {
                             eventManager.publish(new ChannelJoinFailureEvent(name, ChannelJoinFailureEvent.Reason.RETRIES_EXHAUSTED));
                         } else {
-                            log.debug("Chat connection exhausted retries when attempting to join channel: {}", name);
+                            log.warn("Chat connection exhausted retries when attempting to join channel: {}", name);
                         }
                     }
                 })
--- a/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
@@ -1,15 +1,23 @@
 package com.github.twitch4j.chat;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import com.github.benmanes.caffeine.cache.Scheduler;
 import com.github.philippheuer.credentialmanager.CredentialManager;
 import com.github.philippheuer.credentialmanager.domain.OAuth2Credential;
 import com.github.philippheuer.events4j.core.EventManager;
 import com.github.twitch4j.auth.providers.TwitchIdentityProvider;
 import com.github.twitch4j.chat.enums.CommandSource;
 import com.github.twitch4j.chat.enums.TMIConnectionState;
+import com.github.twitch4j.chat.events.AbstractChannelEvent;
 import com.github.twitch4j.chat.events.CommandEvent;
 import com.github.twitch4j.chat.events.IRCEventHandler;
 import com.github.twitch4j.chat.events.channel.ChannelMessageEvent;
+import com.github.twitch4j.chat.events.channel.ChannelNoticeEvent;
+import com.github.twitch4j.chat.events.channel.ChannelStateEvent;
 import com.github.twitch4j.chat.events.channel.IRCMessageEvent;
+import com.github.twitch4j.chat.events.channel.UserStateEvent;
 import com.github.twitch4j.common.config.ProxyConfig;
 import com.github.twitch4j.common.util.CryptoUtils;
 import com.github.twitch4j.common.util.EscapeUtils;
@@ -43,6 +51,7 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
 
 @Slf4j
 public class TwitchChat implements ITwitchChat {
@@ -206,6 +215,16 @@ public class TwitchChat implements ITwitchChat {
     protected final boolean enableMembershipEvents;
 
     /**
+     * The maximum number of attempts to make for joining each channel
+     */
+    protected final int maxJoinRetries;
+
+    /**
+     * Cache of recent number of join attempts for each channel
+     */
+    protected final Cache<String, Integer> joinAttemptsByChannelName;
+
+    /**
      * Helper class to compute delays between connection retries.
      *
      * @see <a href="https://dev.twitch.tv/docs/irc/guide#re-connecting-to-twitch-irc">Official suggestion</a>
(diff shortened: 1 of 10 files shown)


chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -357,6 +388,49 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
...
+        }
+
+        // Remove successfully joined channels from joinAttemptsByChannelName (as further retries are not needed)
+        Consumer<AbstractChannelEvent> joinListener = e -> joinAttemptsByChannelName.invalidate(e.getChannel().getName().toLowerCase());
+        eventManager.onEvent(ChannelStateEvent.class, joinListener::accept);
+        eventManager.onEvent(ChannelNoticeEvent.class, joinListener::accept);
+        eventManager.onEvent(UserStateEvent.class, joinListener::accept);
+
+        // Ban Listener
+        final Set<NoticeTag> banNotices = EnumSet.of(NoticeTag.MSG_BANNED, NoticeTag.MSG_CHANNEL_SUSPENDED, NoticeTag.TOS_BAN); // bit vector
```
iProdigy, line 425: since there's only three states we care about, is this overengineered?

chat/src/main/java/com/github/twitch4j/chat/TwitchChatBuilder.java
```diff
@@ -195,6 +196,21 @@
     @With
     private boolean enableMembershipEvents = true;
 
+    /**
+     * Whether join failures should result in removal from current channels.
+     *
+     * @see ChannelRemovedPostJoinFailureEvent
+     */
+    @With
+    private boolean removeChannelOnJoinFailure = false;
```
iProdigy, line 205: is this too verbose for a variable name

chat/src/main/java/com/github/twitch4j/chat/TwitchChatBuilder.java
```diff
@@ -195,6 +196,21 @@
...
+
+    /**
+     * The maximum number of retries to make for joining each channel, with exponential backoff.
+     * Set to zero or a negative value to disable this feature.
+     */
+    @With
+    private int maxJoinRetries = 3;
```
iProdigy, line 212: this allows 1 (the initial attempt) + 3 (the retries) = 4 join attempts before stopping (until a reconnect, assuming removal isn't enabled) - let me know if you'd prefer a different value

chat/src/main/java/com/github/twitch4j/chat/TwitchChatConnectionPool.java
```diff
@@ -255,6 +256,9 @@ protected TwitchChat createConnection() {
                 .withAutoJoinOwnChannel(false) // user will have to manually send a subscribe call to enable whispers. this avoids duplicating whisper events
         ).build();
 
+        // Reclaim channel headroom upon generic join failures
+        chat.getEventManager().onEvent(threadPrefix + "join-fail-tracker", ChannelRemovedPostJoinFailureEvent.class, e -> unsubscribe(e.getChannelName()));
```
iProdigy, line 260: Should we take `automaticallyPartOnBan` to mean `removeChannelOnJoinFailure` (technically results in parts in more scenarios than just ban)

Or, should we deprecate `automaticallyPartOnBan` in favor of `removeChannelOnJoinFailure`

chat/src/main/java/com/github/twitch4j/chat/events/channel/ChannelRemovedPostJoinFailureEvent.java
```diff
@@ -0,0 +1,49 @@
+package com.github.twitch4j.chat.events.channel;
+
+import lombok.Value;
+
+/**
+ * A chat instance failed to join (or stay joined to) a channel AND removeChannelOnJoinFailure was enabled,
+ * resulting in removal from the instance's current channels.
+ */
+@Value
+public class ChannelRemovedPostJoinFailureEvent {
```
iProdigy, line 10: is this event name too verbose

chat/src/main/java/com/github/twitch4j/chat/TwitchChatConnectionPool.java
PhilippHeuer: I think we can deprecate `automaticallyPartOnBan` in favor of `removeChannelOnJoinFailure`. The reason on why the join failed isn't really important imo, either we want to rejoin or we don't.

chat/src/main/java/com/github/twitch4j/chat/TwitchChatBuilder.java
PhilippHeuer: I think the name is fine because it's obvious what it's doing at a glance.

chat/src/main/java/com/github/twitch4j/chat/TwitchChatBuilder.java
PhilippHeuer: I think a higher default would be better to give it a little more time if there are any weird temporary issues on either twitch or client side. I'd suggest `10`, but we can still tweak the default later.

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
PhilippHeuer: This is a good idea, but we could move this into a static helper class where we could keep similar enum sets for such use cases like first chatter and the user introduction.

chat/src/main/java/com/github/twitch4j/chat/events/channel/ChannelRemovedPostJoinFailureEvent.java
PhilippHeuer: What do you think about naming it `ChannelJoinFailureEvent`?

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -357,6 +391,49 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
...
+                        if (attempts < maxJoinRetries) {
+                            taskExecutor.schedule(() -> {
+                                if (currentChannels.contains(name)) {
+                                    issueJoin(name, attempts + 1);
+                                }
+                            }, initialWait * (1L << (attempts + 1)), TimeUnit.MILLISECONDS); // exponential backoff (pow2 optimization)
+                        } else if (removeChannelOnJoinFailure && removeCurrentChannel(name)) {
+                            eventManager.publish(new ChannelRemovedPostJoinFailureEvent(name, ChannelRemovedPostJoinFailureEvent.Reason.RETRIES_EXHAUSTED));
+                        } else {
+                            log.debug("Chat connection exhausted retries when attempting to join channel: {}", name);
```
PhilippHeuer, line 412: Should we log this as error? I feel like this information is important if the user doesn't handle a `ChannelRemovedPostJoinFailureEvent`.

iProdigy: Yeah I meant to make this at least a `warn`

chat/src/main/java/com/github/twitch4j/chat/TwitchChatBuilder.java
iProdigy: how about 1+7=8, so the maximum wait on default settings is around 10 mins?

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
iProdigy: Sounds good, we can add that class at a later date when there are more enumsets to keep track of

PhilippHeuer (approved):

PR #545, opened 2022-03-04
feat: track chat socket latency

iProdigy:

### Changes Proposed
* Measure time between chat websocket rfc ping/pong and store in `ITwitchChat#getLatency`

### Additional Information
Has been requested by multiple people in the discord

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/ITwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/ITwitchChat.java
@@ -80,6 +80,11 @@ public interface ITwitchChat extends AutoCloseable {
     void close();
 
     /**
+     * @return the most recently measured round-trip latency for the socket(s)
+     */
+    long getLatency();
+
+    /**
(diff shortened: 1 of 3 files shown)


chat/src/main/java/com/github/twitch4j/chat/ITwitchChat.java
```diff
@@ -79,6 +79,11 @@ default boolean sendMessage(String channel, String message, @Unofficial String n
     @Override
     void close();
 
+    /**
+     * @return the most recently measured round-trip latency for the socket(s) in milliseconds, or -1 if unknown
```
iProdigy, line 83: Should a warning be added that this is the *socket* latency, which can be different from the latency of: sending a chat message/command & having it appear/processed by twitch

PhilippHeuer: The description describes what kind of latency this is, so this if fine imo.

PhilippHeuer (approved):

PR #516, opened 2022-01-23
feat: KTX module

SubSide:
### Prerequisites for Code Changes

* [X] This pull request follows the code style of the project
* [X] I have tested this feature

...

### Additional Information

This will add a few kotlin convenience extension functions in a separate library so it can be thrown in if the person wants it.

image

PR #552, opened 2022-03-12
feat: common websocket client

PhilippHeuer:

### Changes Proposed

This improves our testing capabilities in modules that use websocket connections.

* move common code into a websocket-client module
* chat now uses the websocket-client module
* pubSub now uses the websocket-client module
* adds getLatency to pubsub / pubsubpool
* adds a TestEventManager that tracks a list of all published events in order for testing
* uses the spec instead of the builder approach for the WebsocketConnection

and now we can:

* mock the websocket connection and validate that messages would be send as expected
* simulate messages from the websocket server

### Additional Information

This also includes a few example tests for chat / pubSub to show how the usage would look like.
- https://github.com/twitch4j/twitch4j/blob/feature/websocket-common/pubsub/src/test/java/com/github/twitch4j/pubsub/TwitchPubSubTest.java
- https://github.com/twitch4j/twitch4j/blob/feature/websocket-common/chat/src/test/java/com/github/twitch4j/chat/TwitchChatTest.java
- https://github.com/twitch4j/twitch4j/blob/feature/websocket-common/chat/src/test/java/com/github/twitch4j/chat/TwitchChatRateLimitTest.java


client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnection.java
```diff
@@ -0,0 +1,237 @@
...
+            } catch (Exception ex) {
+                log.error("connection to webSocket server {} failed: retrying ...", config.baseUrl(), ex);
+                // Sleep before trying to reconnect
+                try {
+                    config.backoffStrategy().sleep();
+                } catch (Exception ignored) {
+
+                } finally {
+                    // reconnect
+                    reconnect();
```
stachu540, line 171: Can we adding some connection retry count (in configuration) when should not do any force connection in `n` retries? In default `-1` when will be a loop connection to retrying until killing machine?

PhilippHeuer: delays between connection attempts are currently handled by the backoffStrategy a few lines higher `config.backoffStrategy().sleep();`

https://github.com/twitch4j/twitch4j/blob/feature/websocket-common/client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java#L49-L58

stachu540: didn't add about it stops after `n` count retries.

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
```diff
@@ -0,0 +1,101 @@
...
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import org.apache.commons.lang3.RandomStringUtils;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.function.Consumer;
+
+@Accessors(chain = true, fluent = true, prefix = "")
```
iProdigy, line 15: With `fluent=true`:
```kotlin
WebsocketConnection() {
	it.baseUrl("https://twitch4j.github.io")
	it.wsPingPeriod(1337)
	it.onTextMessage { msg -> println(msg) }
}
```

&nbsp;

With `fluent=false`:
```kotlin
WebsocketConnection() {
	it.baseUrl = "https://twitch4j.github.io"
	it.wsPingPeriod = 1337
	it.onTextMessage = Consumer { msg -> println(msg) }
}
```

&nbsp;

I personally find the latter kotlin syntax to be more idiomatic, but this makes the event handlers feel clunky in java (e.g., `getOnTextMessage`, `setOnTextMessage`), so no change is really needed here. Since end users are unlikely to directly interface with this class, this is more a general comment on how we should approach other specs...

Even more idiomatic would be to remove the `it.` (e.g., so it's just `wsPingPeriod = 1337`) — but this may be a task for the kotlin module to achieve (via extensions functions + lambdas with receiver to create a type-safe builder)

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
```diff
@@ -0,0 +1,101 @@
...
+        data.validate();
+        return data;
+    }
+
+    /**
+     * validate the config
+     */
+    public void validate() {
+        Objects.requireNonNull(baseUrl, "baseUrl is a required parameter!");
+        Objects.requireNonNull(wsPingPeriod, "wsPingPeriod is a required parameter, set to 0 to disable!");
```
iProdigy, line 32: primitive will never fail this check

iProdigy (commented):
Thanks for getting this off the ground!

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * called after the disconnect
+     */
+    private Runnable onPostDisconnect;
+
+    private String proxyHost;
+    private Integer proxyPort;
+    private String proxyUsername;
+    private String proxyPassword;
```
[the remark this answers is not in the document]
PhilippHeuer: I tried to make `client-websocket` fully standalone. Right now the only reason it still depends on `twitch4j-common` is that i wasn't sure what do do about the `backoffStrategy` yet - but if we keep the dep on common we can also use the `ProxyConfig`.

chat/src/main/java/com/github/twitch4j/chat/enums/TMIConnectionState.java
```diff
@@ -1,12 +0,0 @@
-package com.github.twitch4j.chat.enums;
-
-/**
- * WebSocket - Connection State
- */
-public enum TMIConnectionState {
```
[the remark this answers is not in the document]
iProdigy: Addressed 👍 https://github.com/twitch4j/twitch4j/pull/552/commits/a7dcf04e4c01cb1ab73e6128ac6664a8b4337a17#diff-d96c95d3eba0e6ca5afe8b01bb69a62391fab0206554706fa8b115d33f552f00R799

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
PhilippHeuer: I also prefer `fluent=true` because we can improve kotlin usage with the kotlin module later, but we can't do that for java.  

We could add the kotlin extension function in a follow up PR after #516 is merged. This isn't really required for this spec as this is internal, but if we want to provide the improved usage from kotlin i think it's important to do it for all specs (also in the codegen for helix/... later).

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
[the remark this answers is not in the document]
iProdigy: To go standalone, we could:

* Create `IBackoffStrategy` interface in the websocket module (would have methods `get()`, `reset()`, `sleep()`)
* Remove explicit references of `ExponentialBackoffStrategy` in `WebsocketConnection`/`WebsocketConnectionConfig` (note: this will require people to pass a concrete implementation when making the config so the backoff strategy is not null)
* Create an adapter class that implements `IBackoffStrategy` & takes `ExponentialBackoffStrategy` in the constructor


As a result, usage would look like:
```kotlin
WebsocketConnection() {
	it.baseUrl("https://twitch4j.github.io")
	it.backoffStrategy(ExponentialBackoffAdapter(ExponentialBackoffStrategy.builder().build()))
}
```


Up to you if this route is desirable

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnection.java
iProdigy: `ExponentialBackoffStrategy` actually has a `maxRetries` field that can be configured. however's it's not currently honored by the socket; if `config.backoffStrategy().sleep()` returns false, we should not attempt reconnect

iProdigy: Fixed bug in https://github.com/twitch4j/twitch4j/pull/552/commits/bc4bfea28a400b1c6736943fe9cacdbe49fb4607 - though I don't think users want `n` to be non-negative by default (so it will retry forever, unless user configures otherwise)

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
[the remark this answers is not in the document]
PhilippHeuer: I'd still like to make this module useable standalone, even outside of t4j.

I was thinking along the lines of moving some of the useful classes (`ProxyConfig`, `ExponentialBackoffStrategy`, ...) into a `twitch4j-utils` module that has no dependencies itself. I just don't want to use the current common because it includes dependencies (auth, events4j, jackson and other stuff) that shouldn't end up in the websocket client.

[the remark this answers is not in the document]
iProdigy: utils module + interface sounds good!

Shall we directly depend on the classes for now & move them on the next major version bump?
Or could we get away with making common depend on utils module (right now) & move the classes such that the package names are the same (so we virtually avoid the breaking change for most folks)

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnection.java
```diff
@@ -0,0 +1,243 @@
...
+                    lastPing.compareAndSet(0L, System.currentTimeMillis());
+                }
+            }
+
+            @Override
+            public void onPongFrame(WebSocket websocket, WebSocketFrame frame) {
+                final long last = lastPing.getAndSet(0L);
+                if (last > 0) {
+                    latency = System.currentTimeMillis() - last;
+                    log.trace("TwitchChat: Round-trip socket latency recorded at {} ms.", latency);
```
iProdigy, line 134:
```suggestion
                    log.trace("T4J Websocket: Round-trip socket latency recorded at {} ms.", latency);
```

client-websocket/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+// In this section you declare the dependencies for your production and test code
+dependencies {
+	// Common
+	api(project(":twitch4j-util"))
+
+	// WebSocket
+	api(group = "com.neovisionaries", name = "nv-websocket-client")
```
iProdigy, line 7: Since the details of the websocket implementation are not exposed to the user, we *could* remove `api` here (which is one less thing to worry about if we ever decide to switch ws libraries)
```suggestion
	implementation(group = "com.neovisionaries", name = "nv-websocket-client")
```

chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -314,18 +259,27 @@ public TwitchChat(EventManager eventManager, CredentialManager credentialManager
...
-        this.wsPingPeriod = wsPingPeriod;
 
-        // Create WebSocketFactory and apply proxy settings
-        this.webSocketFactory = new WebSocketFactory();
-        if (proxyConfig != null)
-            proxyConfig.applyWs(webSocketFactory.getProxySettings());
+        // init connection
+        if (websocketConnection == null) {
+            this.connection = new WebsocketConnection(spec -> {
+                spec.baseUrl(baseUrl);
```
iProdigy, line 266: Should we allow `IExponentialStrategy` to be set in chat/pubsub builder so users can override details like `maxRetries` (and we'll pass it along via the spec)

pubsub/src/main/java/com/github/twitch4j/pubsub/TwitchPubSubBuilder.java
```diff
@@ -61,6 +71,12 @@ public class TwitchPubSubBuilder {
     @With
     private int wsPingPeriod = 15_000;
 
+    /**
+     * WebSocket Connection Backoff Strategy
+     */
+    @With
+    private IBackoffStrategy connectionBackoffStrategy = null;
```
iProdigy, line 78: This technically changes the default backoff settings for pubsub from [master](https://github.com/twitch4j/twitch4j/blob/af5744a5a4e142f443cb8ccca1d54163d60c5079/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java#L250-L256), but is relatively minor so this is fine 👍

client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnectionConfig.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Task Executor
+     */
+    private ScheduledExecutorService taskExecutor = new ScheduledThreadPoolExecutor(2);
+
+    /**
+     * Helper class to compute delays between connection retries
+     */
+    private IBackoffStrategy backoffStrategy = ExponentialBackoffStrategy.builder()
+        .immediateFirst(true)
```
iProdigy, line 73:
```suggestion
        .immediateFirst(false)
```
With the pubsub change below and connection pools relying on this, perhaps we delay the first reconnect to avoid a thundering herd issue

iProdigy (approved):
Great work on this!

PR #561, opened 2022-04-16
feat: comply with undocumented helix rate limits

iProdigy:

### Issues Fixed 
* Prevents 429's from some helix endpoints with special undocumented rate limits
* Prevents 429's from consistently hammering (general) helix endpoints (due to refill rate asymmetry)

### Changes Proposed
* Restrict `banUser`/`unbanUser` calls to 100/30 bucket per channel
* Restrict `addBlockedTerm`/`removeBlockedTerm` calls to 60/60 bucket per channel
* Restrict `createClip` calls to 600/60 bucket per user
* Reduce helix bucket *refill* rate to 600/60
* Allow users to override the global helix ratelimit (to be less than 800/60, for example)
* Fix resource leak in error decoders

### DIFF
--- a/rest-helix/src/main/java/com/github/twitch4j/helix/interceptor/TwitchHelixClientIdInterceptor.java
+++ b/rest-helix/src/main/java/com/github/twitch4j/helix/interceptor/TwitchHelixClientIdInterceptor.java
@@ -19,6 +19,7 @@ import org.apache.commons.lang3.StringUtils;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 /**
  * Injects ClientId Header, the User Agent and other common headers into each API Request
@@ -34,6 +35,12 @@ public class TwitchHelixClientIdInterceptor implements RequestInterceptor {
      */
     @Unofficial
     private static final Bandwidth BANS_BANDWIDTH = Bandwidth.simple(100, Duration.ofSeconds(30));
+
+    /**
+     * Empirically determined rate limit on the helix create clip endpoint, per user
+     */
+    @Unofficial
+    private static final Bandwidth CLIPS_BANDWIDTH = Bandwidth.simple(600, Duration.ofSeconds(60));
 
     /**
      * Reference to the Client Builder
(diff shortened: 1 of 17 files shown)


rest-helix/src/main/java/com/github/twitch4j/helix/interceptor/TwitchHelixDecoder.java
```diff
@@ -27,11 +28,24 @@ public Object decode(Response response, Type type) throws IOException {
...
+
+            // Synchronize library buckets with twitch data
             if (remaining != null) {
-                try {
-                    interceptor.updateRemaining(token.substring(BEARER_PREFIX.length()), Integer.parseInt(remaining));
-                } catch (Exception ignored) {
+                String bearer = token.substring(BEARER_PREFIX.length());
+                if (response.request().httpMethod() == Request.HttpMethod.POST && response.request().requestTemplate().path().endsWith("/clips")) {
+                    // Create Clip has a separate rate limit to synchronize
+                    interceptor.updateRemainingCreateClip(bearer, remaining);
```
iProdigy, line 45: [Create Clip](https://dev.twitch.tv/docs/api/reference#create-clip) is supposed to include `Ratelimit-Helixclipscreation-Limit` & `Ratelimit-Helixclipscreation-Remaining` headers - when testing, these headers weren't present, and twitch simply used the standard rate limit header to express the value (so we simply use `remaining`)

Also worth noting: the moderation APIs with undocumented rate limits do not have any special header behavior, so there is no analogous `updateRemaining` call (as we can't determine endpoint-specific bucket size/remaining)

rest-helix/src/main/java/com/github/twitch4j/helix/interceptor/TwitchHelixClientIdInterceptor.java
```diff
@@ -123,6 +79,7 @@ public TwitchHelixClientIdInterceptor(TwitchHelixBuilder twitchHelixBuilder) {
         this.defaultClientId = twitchAPIBuilder.getClientId();
         this.apiRateLimit = twitchAPIBuilder.getApiRateLimit();
         this.defaultAuthToken = twitchHelixBuilder.getDefaultAuthToken();
+        this.rateLimitTracker = new TwitchHelixRateLimitTracker(this);
```
iProdigy, line 82: Should this be created in `TwitchHelixBuilder` instead, and passed in the constructor to each class that needs access?

PhilippHeuer: This makes injecting the meterRegistry a lot easier when we add metric-support to monitor bucket-usage in the future - so this is a good idea.

PhilippHeuer (approved):
Good job - and great investigation on the 429's :)

PR #566, opened 2022-05-05
feat: support experimental direct cheer usernotice

iProdigy:

### Changes Proposed
* Parse unofficial `midnightsquid` usernotice and fire `DirectCheerEvent`

### Additional Information
https://help.twitch.tv/s/article/cheering-experiment-2022?language=en_US

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/events/channel/DirectCheerEvent.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/events/channel/DirectCheerEvent.java
@@ -1,0 +1,137 @@
+package com.github.twitch4j.chat.events.channel;
+
+import com.github.twitch4j.common.annotation.Unofficial;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Value;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.math.BigDecimal;
+import java.util.Collections;
+import java.util.Currency;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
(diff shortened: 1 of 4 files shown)


chat/src/main/java/com/github/twitch4j/chat/events/channel/DirectCheerEvent.java
```diff
@@ -0,0 +1,137 @@
...
+    public DirectCheerEvent(IRCMessageEvent event) {
+        this(
+            event,
+            event.getTagValue("msg-param-amount").map(Integer::parseInt).orElse(null),
+            event.getTagValue("system-msg").map(String::trim).orElse("")
+        );
+    }
+
+    DirectCheerEvent(IRCMessageEvent event, Integer amount, @NotNull String systemMessage) {
+        super(event, event.getChannel(), event.getUser(), event.getMessage().orElse(""), Math.round(amount * CREATOR_REVENUE_SPLIT), event.getSubscriberMonths().orElse(0), event.getSubscriptionTier().orElse(0), event.getFlags());
```
PhilippHeuer, line 87: Just to be save, should we make sure that `amount` arg can not be null because of `Math.round(amount * CREATOR_REVENUE_SPLIT)` ? Just in case they change sth. up as it's still expermental.

Maybe by checking that `msg-param-amount` is present on not null in IRCEventHandler#onDirectCheer? (or default to 0 instead of null in line 81)

iProdigy: good point, would agree with defaulting to `0` or `-1` (maybe `0` is best in case people are tracking tips over time & forget to check sign & tip amount falsely decreases marginally)

PhilippHeuer (approved):

PR #576, opened 2022-05-23
feat: implement send extension message rate limit

iProdigy:

### Related Issues
* https://github.com/twitchdev/issues/issues/612

### Changes Proposed
* Add endpoint-specific rate limiting for `sendExtensionChatMessage`/`sendExtensionPubSubMessage`


rest-helix/src/main/java/com/github/twitch4j/helix/interceptor/TwitchHelixHttpClient.java
```diff
@@ -99,10 +100,47 @@ private Response delegatedExecute(Request request, Request.Options options) thro
...
+            // Conform to endpoint-specific bucket
+            Bucket pubSubBucket = rateLimitTracker.getExtensionPubSubBucket(clientId, target);
+            return executeAgainstBucket(pubSubBucket, () -> client.execute(request, options));
+        }
+
         // no endpoint-specific rate limiting was needed; simply perform network request now
         return client.execute(request, options);
     }
 
+    static String getExtensionPubSubTarget(byte[] body) {
```
iProdigy, line 129: does this belong in a different class

iProdigy: a possible alternative approach: add the target as an extra query param (that we make up) & hope twitch ignores it

PhilippHeuer: Would putting it into a header field and then taking it out of there be a option (easier to access than get-params)?
We could also use `request.headers().remove("KEY");` to get rid of the header before the request is sent in here.

iProdigy: Implemented the header approach, but didn't remove the header because of the bucket update post-request

PhilippHeuer (approved):

PR #597, opened 2022-07-07
feat: emit websocket events

samfundev:

### Changes Proposed

* Websockets now emit events when their state changes
* Websockets have a new state when their connection is lost


iProdigy:
what's the use case in mind for `WebsocketStateEvent`?

samfundev:
> what's the use case in mind for `WebsocketStateEvent`?

I'll be using it to show the status of the connection to chat. For example, if the connection gets disconnected that could be displayed to the user.

client-websocket/src/main/java/com/github/twitch4j/client/websocket/events/WebsocketStateEvent.java
```diff
@@ -0,0 +1,27 @@
...
+import lombok.EqualsAndHashCode;
+
+/**
+ * Websocket State Event
+ * <p>
+ * Called when the websocket changes its connection state.
+ */
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class WebsocketStateEvent extends Event {
```
iProdigy, line 15: Since events4j isn't meant to be referenced in [this module](https://github.com/twitch4j/twitch4j/blob/master/client-websocket/build.gradle.kts), perhaps the chat & pubsub modules should have their own module-specific event (& this would avoid the annoying cast from `AutoCloseable`)

PhilippHeuer (approved):
LGTM, thanks for the contribution.

iProdigy (approved):
Thanks for your contribution!

PR #610, opened 2022-07-22
feat: add fetch rewards via gql

iProdigy:

### Changes Proposed
* Add `TwitchGraphQL#fetchChannelPointRewards`

### DIFF
--- a/graphql/src/main/graphql/com/github/twitch4j/graphql/internal/fetchChannelPointRewards.graphql
+++ b/graphql/src/main/graphql/com/github/twitch4j/graphql/internal/fetchChannelPointRewards.graphql
@@ -1,0 +1,138 @@
+query fetchChannelPointRewards($login: String!) {
+	channel(name: $login) {
+		id
+		communityPointsSettings {
+			name
+			image {
+				url
+				url2x
+				url4x
+			}
+			isEnabled
+			automaticRewards {
+				backgroundColor
+				cost
+				defaultBackgroundColor
+				defaultCost
+				defaultImage {
+					url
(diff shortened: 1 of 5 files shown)


graphql/src/main/graphql/com/github/twitch4j/graphql/internal/fetchChannelPointRewards.graphql
```diff
@@ -0,0 +1,138 @@
+query fetchChannelPointRewards($login: String!) {
```
iProdigy, line 1: this technically should be `fetchCommunityPointsSettings`, but don't think that's clear enough to end users that it would contain the custom reward list

graphql/src/main/java/com/github/twitch4j/graphql/TwitchGraphQL.java
```diff
@@ -187,6 +187,10 @@ public CommandUpdateClip updateClip(OAuth2Credential auth, String slug, String n
         return new CommandUpdateClip(getApolloClient(auth), slug, newTitle);
     }
 
+    public CommandFetchChannelPointRewards fetchChannelPointRewards(OAuth2Credential auth, String channelLogin) {
```
PhilippHeuer, line 190: should we add a comment here referencing the real operation name `communityPointsSettings `?

iProdigy: update the class name to reflect the real operation name; better now?

PhilippHeuer (approved):

PR #618, opened 2022-08-05
refactor: use xanthic cache facade

iProdigy:

### Issues Fixed 
* Forced Caffeine use was incompatible with Android

### Changes Proposed
* Switch all direct Caffeine use to the Xanthic facade

### Additional Information
For now, we still depend on `cache-provider-caffeine` to avoid any breaking changes... users can exclude this dependency & specify a different provider


PhilippHeuer:
Should we also allow to explicitly set a provider in the builder?

iProdigy:
> Should we also allow to explicitly set a provider in the builder?

We have some `static final` caches where it'd be tricky to pass that state to (e.g., https://github.com/twitch4j/twitch4j/blob/refactor/xanthic-cache/eventsub-common/src/main/java/com/github/twitch4j/eventsub/util/EventSubVerifier.java#L35)

So users can just do `CacheApiSettings#setDefaultCacheProvider` instead

PhilippHeuer (approved):

PR #627, opened 2022-08-25
feat: allow custom websocket close delay

iProdigy:

### Issues Fixed 
* Allows for faster underlying socket close for applications where this is required (e.g., minecraft plugins on disable, which triggers class unloading)

### Changes Proposed
* Add `closeDelay` to `WebsocketConnectionConfig`
* Don't return from `WebsocketConnection#close` until the underlying `Socket` has fully closed

### Additional Information
`nv-websocket-client` defaults `closeDelay` to `10_000` - this seems high to just wait for a single frame, so we default to `1000` instead (this part of the websocket protocol is just good manners anyway, so it's not the end of the world if formal close isn't received)

### DIFF
--- a/client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnection.java
+++ b/client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnection.java
@@ -308,7 +308,7 @@ public class WebsocketConnection implements AutoCloseable {
         } finally {
             // await the close of the underlying socket
             try {
-                boolean completed = closeLatch.await(Math.max(1000, config.closeDelay()) * 2L, TimeUnit.MILLISECONDS);
+                boolean completed = closeLatch.await(config.closeDelay() + 1000L, TimeUnit.MILLISECONDS);
                 if (completed) {
                     log.trace("Underlying websocket complete close was successful");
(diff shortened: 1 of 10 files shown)


client-websocket/src/main/java/com/github/twitch4j/client/websocket/WebsocketConnection.java
```diff
@@ -172,6 +182,9 @@ protected void setState(WebsocketConnectionState newState) {
     public void connect() {
         WebsocketConnectionState connectionState = this.connectionState.get();
         if (connectionState == WebsocketConnectionState.DISCONNECTED || connectionState == WebsocketConnectionState.RECONNECTING || connectionState == WebsocketConnectionState.LOST) {
+            if (closed.get())
+                throw new IllegalStateException("WebsocketConnection was already closed!");
```
iProdigy, line 186: improbable breaking change: `connect()` can no longer be called after `close()`

PhilippHeuer: the instance shouldn't be usable after calling `close()`, so adding this is good

PhilippHeuer (approved):

PR #638, opened 2022-09-03
perf: avoid regex split on chat message

iProdigy:

### Changes Proposed
* Switches `String#split` (regex-based) to `StringUtils#split` (non-regex) for better performance in `TwitchChat#onTextMessage`

### Additional Information
Achieved 9.5% faster parsing in my testing

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
@@ -530,8 +530,7 @@ public class TwitchChat implements ITwitchChat {
     }
 
     protected void onTextMessage(String text) {
-        Arrays.asList(text.replace("\n\r", "\n")
-                .replace("\r", "\n").split("\n"))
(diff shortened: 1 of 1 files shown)


chat/src/main/java/com/github/twitch4j/chat/TwitchChat.java
```diff
@@ -761,6 +760,10 @@ private boolean removeCurrentChannel(String channelName) {
     @Override
     @SuppressWarnings("ConstantConditions")
     public boolean sendMessage(String channel, String message, Map<String, Object> tags) {
+        if (message.length() > 500) {
```
PhilippHeuer, line 763: the message won't be truncated to length 500 on server side so a longer message would just disappear.
So we could either truncate longer messages to length 500 ourselves or we directly return after the warning (without a send attempt)

iProdigy: good point, I initially didn't want to interfere in case there was a valid reason to exceed 500, but those users can just do `sendRaw` anyways

PhilippHeuer (approved):

PR #656, opened 2022-09-21
fix: support named capture groups on Android

samfundev:

### Changes Proposed

* Add a library which supports named capture groups.

### Additional Information

Android doesn't support named capture groups [before API 26](https://developer.android.com/reference/java/util/regex/Pattern#group-name).

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
@@ -13,8 +13,8 @@ import lombok.*;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.*;
-import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


chat/src/main/java/com/github/twitch4j/chat/events/channel/IRCMessageEvent.java
```diff
@@ -13,8 +13,8 @@
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.*;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import com.google.code.regexp.Pattern;
+import com.google.code.regexp.Matcher;
```
Felanbird, line 17: why did Matcher and Pattern get re-ordered?

iProdigy (approved):
Benchmarks look good - thanks for the contribution!

PhilippHeuer (approved):

PR #667, opened 2022-10-01
fix: handle new direct cheer tags for currency

iProdigy:

### Changes Proposed
* Handle new `msg-param-exponent` & `msg-param-currency` tags in `DirectCheerEvent`

### Additional Information
* Round 2 of `midnightsquid` allows for non-USD direct cheers
* If the currency is not USD, we don't provide a corresponding bits value (0), as we don't keep track of currency exchange rates

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/events/channel/DirectCheerEvent.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/events/channel/DirectCheerEvent.java
@@ -1,18 +1,12 @@
 package com.github.twitch4j.chat.events.channel;
 
 import com.github.twitch4j.common.annotation.Unofficial;
+import com.github.twitch4j.common.util.DonationAmount;
 import lombok.EqualsAndHashCode;
-import lombok.Getter;
 import lombok.Value;
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
 import java.math.BigDecimal;
-import java.util.Collections;
 import java.util.Currency;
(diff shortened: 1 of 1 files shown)


chat/src/main/java/com/github/twitch4j/chat/events/channel/DirectCheerEvent.java
```diff
@@ -78,60 +38,51 @@ public class DirectCheerEvent extends CheerEvent {
...
-    DirectCheerEvent(IRCMessageEvent event, Integer amount, @NotNull String systemMessage) {
-        super(event, event.getChannel(), event.getUser(), event.getMessage().orElse(""), Math.round(amount * CREATOR_REVENUE_SPLIT), event.getSubscriberMonths().orElse(0), event.getSubscriptionTier().orElse(0), event.getFlags());
-        this.amount = amount;
-        this.systemMessage = systemMessage;
+    DirectCheerEvent(IRCMessageEvent event, long amount, int exponent, String currency) {
+        super(event,
+            event.getChannel(),
+            event.getUser(),
+            event.getMessage().orElse(""),
+            "USD".equalsIgnoreCase(currency) ? Math.round(amount * CREATOR_REVENUE_SPLIT) : 0,
```
iProdigy, line 52: The `#getBits` method can be deprecated now to discourage use as we don't track currency fluctuations or payment processor fees

If bits as a whole are killed by twitch (increasingly likely), the whole super class can be deprecated

PhilippHeuer (approved):

PR #698, opened 2023-01-05
Improve build cacheability

jaronsummers:

### Changes Proposed

- Explicitly declare places where manifest is a dependency
- normalize runtime classpath analysis of manifest

### Additional Information

These changes resolved most parts of the build that Gradle refused to build incrementally, or cache. On my local machine, the time for repeated `check` executions went from 13 seconds to 0.7 seconds.

If you have any questions/concerns, let me know!

Note: I did not check the contributing guidelines because the link in the readme goes to a dead link, so if there are any issues with style or formatting, I'm happy to fix them.

Here's a build scan of the current master (second build): https://scans.gradle.com/s/esrc625h6w7gw
Here's a build scan of this fork (also second build): https://scans.gradle.com/s/dxxpiwazvenyi

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,6 +59,10 @@ subprojects {
 	apply(plugin = "com.coditory.manifest")
 	apply(plugin = "com.github.johnrengelman.shadow")
 
+	manifest {
+		buildAttributes = false
+	}
+
 	lombok {
 		version.set("1.18.24")
 		disableConfig.set(true)


jaronsummers:
I corrected the merge conflict after the branch change, I haven't re-tested with the new dependency versions, but I don't think there would be any issues.

iProdigy:
Thank you for the contribution!

&nbsp;

Could this also be achieved by the [following](https://github.com/coditory/gradle-manifest-plugin#overriding-generated-attributes)?
```kt
subprojects {
  manifest {
    buildAttributes = false
  }
}
```

jaronsummers:
Yep, if there's nothing else in the project depending on that metadata, your suggestion is great.

I'll set aside some time tomorrow morning to test & update my PR.

jaronsummers:
Yep, I had the same findings as you: it works great.

This approach does result in the manifest being regenerated on every build, because it doesn't declare outputs, but that operation only takes a fraction of a second on my machine, and in my brief experimentation, declaring the outputs in the obvious/trivial way caused significant new pain, so I think if you are happy with this, so am I. :)

PR #715, opened 2023-01-31
chore: improve gradle caching

stepango:

### Issues Fixed 
* `io.freefair.lombok`, `com.coditory.manifest`, `com.github.johnrengelman.shadow` plugins applied to the root project, but never used
*  `javadoc` task configured for root project, but never used
* `jar` and `javadoc` tasks implicitly depend on `manifest` output which breaks the build cache
* `manifest` task from `com.coditory.manifest` plugin implemented incorrectly(does not declare outputs) which caused all dependent tasks to execute
* `manifest` task produces different output every time it's executed
* `javadoc` task configured with the absolute path for `options.overview` arg, which prevents cache from being reused in case of project relocation(moved to different dir, or reuse of build-cache generated by CI)

### Changes Proposed

* Use `apply(false)` in root project plugins block - speeds up project configuration by avoiding tasks creation in root project, please note, this approach can't be used for default Gradle plugins. 
* `javadoc` task configuration moved from `allprojects` to `subprojects` block
* Added explicit dependencies of `jar` and `javadoc` tasks to `manifest` task
* Modified `manifest` task with declared outputs
* `javadoc.options.overview` now uses relative path to allow caching
* manifest plugin only enabled when either condition is met
   * `-PenableManifest` is supplied via cli
   *  `publish` task is executed

### Additional Information

You can find attached reports below.

### Before:
Build scan link: https://scans.gradle.com/s/7w3yauarjdv6o/performance/execution
Screenshot 2023-01-31 at 2 20 00 PM

### After:
https://scans.gradle.com/s/htzrk3pymj4ka/performance/execution
Screenshot 2023-01-31 at 2 19 52 PM

The results of this experiment could be replicated using https://scans.gradle.com/

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,4 @@
 import com.coditory.gradle.manifest.ManifestPluginExtension
-import com.coditory.gradle.manifest.ManifestTask
 import com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import io.freefair.gradle.plugins.lombok.LombokExtension
@@ -25,25 +24,22 @@ allprojects {
 	}
 }
 
+val enableManifest = project.gradle.startParameter.taskNames
+		.any { s -> s.startsWith("publish") }
+
 // Subprojects
 subprojects {
 	apply(plugin = "signing")
 	apply(plugin = "java-library")
 	apply(plugin = "maven-publish")
 	apply(plugin = "io.freefair.lombok")
-	apply(plugin = "com.coditory.manifest")
 	apply(plugin = "com.github.johnrengelman.shadow")
 
-	project.extensions.getByType(ManifestPluginExtension::class.java).apply {
-		buildAttributes = false
-	}
-
-	normalization {
-		runtimeClasspath {
-			metaInf {
-				ignoreManifest()
-			}
-		}
+	if (enableManifest) {
+		apply(plugin = "com.coditory.manifest")
+		project.extensions
+				.getByType(ManifestPluginExtension::class.java)
+				.apply { buildAttributes = false }
 	}
 
 	project.extensions.getByType(LombokExtension::class).apply {
@@ -158,10 +154,6 @@ subprojects {
 	// Source encoding
 	tasks {
 
-		withType<ManifestTask>().configureEach {
-			outputs.file(File(buildDir, "resources/main/META-INF/MANIFEST.MF"))
-					.withPropertyName("outputFile")
-		}
 		// shadowjar & relocation
 		val relocateShadowJar by creating(ConfigureShadowRelocation::class) {
 			target = named<ShadowJar>("shadowJar").get()
@@ -174,9 +166,9 @@ subprojects {
 				dependsOn(relocateShadowJar)
 				archiveClassifier.set("shaded")
 			}
-			val manifestTask = named("manifest")
-			dependsOn(manifestTask)
-			manifest.from(manifestTask.map { it.outputs.files.first() })
+			if (enableManifest) {
+				manifest.from(File(buildDir, "resources/main/META-INF/MANIFEST.MF"))
+			}
 		}
 
 		// compile options
@@ -216,7 +208,7 @@ subprojects {
 		// javadoc & delombok
 		val delombok by getting(Delombok::class)
 		javadoc {
-			dependsOn(delombok, named("manifest"))
+			dependsOn(delombok)
 			source(delombok)
 			options {
 				title = "${project.name} (v${project.version})"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,52 +1,26 @@
+import com.coditory.gradle.manifest.ManifestTask
+import com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -59,11 +34,19 @@ subprojects {
 	apply(plugin = "com.coditory.manifest")
 	apply(plugin = "com.github.johnrengelman.shadow")
 
-	manifest {
+	project.extensions.getByType(ManifestPluginExtension::class.java).apply {
```
runningcode, line 37: The syntax used before is cleaner and easier to understand. Why the change? (same for the `LombokExtension`.

build.gradle.kts
```diff
@@ -59,11 +34,19 @@ subprojects {
...
 		buildAttributes = false
 	}
 
-	lombok {
+	normalization {
+		runtimeClasspath {
+			metaInf {
+				ignoreManifest()
```
runningcode, line 44: Is this still needed given that line 38 disables the `buildAttributes` in the manifest?

build.gradle.kts
```diff
@@ -174,37 +157,73 @@ subprojects {
...
+		withType<ManifestTask> {
+			outputs.file(File(buildDir, "resources/main/META-INF/MANIFEST.MF"))
+					.withPropertyName("outputFile")
+
+		}
 		// shadowjar & relocation
-		val relocateShadowJar by creating(com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation::class) {
-			target = shadowJar.get()
+		val relocateShadowJar by creating(ConfigureShadowRelocation::class) {
+			target = named<ShadowJar>("shadowJar").get()
```
runningcode, line 168: I would remove this change. Using the lazy `named<ShadowJar>("shadowJar")` is redundant if immediately call `.get()` on the provider.

build.gradle.kts
```diff
@@ -174,37 +157,73 @@ subprojects {
 
 	// Source encoding
 	tasks {
+
+		withType<ManifestTask> {
```
runningcode, line 161:
```suggestion
		withType<ManifestTask>().configureEach {
```
We should use the `.configureEach` here in order to use a lazy configuration.

Separate question:
Now that we are adding an output, this task will have an up-to-date checking mechanism. Before we allow that, are we sure that _all_ inputs and _all_ outputs for this ManifestTask are properly declared?

runningcode (commented):
Looks good, I had just some questions about specific changes.

build.gradle.kts
stepango: We are not applying plugins for the root project, so synthetic properties e.g. `manifest` is not available.

build.gradle.kts
stepango: same reason as above, `shadowJar` property is not available if plugin is not applied

runningcode: ah makes sense then

build.gradle.kts
runningcode: oh i see. the time stamp is no good for reproducibility.

build.gradle.kts
runningcode: I don't see where any of the inputs such as group, version, scm, date, etc are tracked as inputs here. You mentioned that they were inputs in this comment: https://github.com/twitch4j/twitch4j/pull/715#discussion_r1093433608

Without that, I recommend not declaring the output directory for now.

build.gradle.kts
runningcode: Makes sense, thanks

build.gradle.kts
stepango: Yep, this data is not tracked as inputs.
https://github.com/coditory/gradle-manifest-plugin/blob/master/src/main/kotlin/com/coditory/gradle/manifest/ManifestAttributes.kt

It's parsed from git commands output and system properties.

Removing the output directory declaration will cause `javadoc` and `jar` tasks to fail because of the detected undeclared dependency of the metadata file.
`manifest` task
Screenshot 2023-02-01 at 9 48 10 AM
`jar task`
Screenshot 2023-02-01 at 9 49 04 AM

stepango: This said, one of the options I see is to disable the manifest plugin for all use cases except for publishing. 
This is an idea pretty similar to what Android does by default - having separate `debug` and `release` configs for the build.
e.g. 
```kotlin
if (project.properties.containsKey("enableManifest")) 
  apply(plugin = "com.coditory.manifest")
```
```shell
./gradlew publish -PenableManifest
```

@iProdigy what do you think?

PhilippHeuer: @stepango thanks for the pull request and the investigation :)

Disabling the plugin outside of `publish` seems like a good idea. (until the plugin is fixed)
Would the following change work to automate the detection without a new property?

`project.properties.containsKey("enableManifest")`
->
`project.gradle.startParameter.taskNames.any { s -> s.startsWith("publish") }`

stepango: This will work as long as you are not using short names for publish task e.g. `pub` :)

stepango: I've ended up adding both options
``` kotlin
project.properties.containsKey("enableManifest") ||
  project.gradle.startParameter.taskNames.any { s -> s.startsWith("publish") }
```

I've kept the flag for testing purposes, e.g. you want to see what is inside the manifest files without actually publishing 

@iProdigy @PhilippHeuer let me know if this works for you.

PhilippHeuer (approved):

iProdigy (approved):
Thank you for the contribution! Huge caching improvements https://scans.gradle.com/s/tsnggbalnrxhu => https://scans.gradle.com/s/54bj3v5wcrplc

PR #819, opened 2023-07-20
fix: npe when parsing global user state in channelName lookup

PhilippHeuer:

### Changes Proposed

* null-check on channelId before channel name lookup
* use `ConcurrentHashMap` in parse test-cases to ensure identical behavior

### Additional Information

Our test case didn't catch this because `Collections.emptyMap` allows null-key lookups while the `ConcurrentHashMap` used in the real code doesn't / throws a NPE.


iProdigy (approved):
Good catch!

PR #821, opened 2023-07-22
perf: split irc messages more efficiently

iProdigy:

### Changes Proposed
* Reduce string/array allocations in `TwitchChat#onTextMessage(String)` via `MessageParser.consumeLines`

### DIFF
--- a/chat/src/main/java/com/github/twitch4j/chat/util/MessageParser.java
+++ b/chat/src/main/java/com/github/twitch4j/chat/util/MessageParser.java
@@ -132,21 +132,18 @@ public class MessageParser {
 
     @ApiStatus.Internal
     public void consumeLines(@NotNull String source, @NotNull Consumer<String> consumer) {
+        int len = source.length();
         int start = 0;
         int i;
-        while (true) {
-            i = source.indexOf('\n', start);
-            if (i < 0) {
-                if (start < source.length()) {
-                    consumer.accept(source.substring(start));
-                }
-                break;
-            } else {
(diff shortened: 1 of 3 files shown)


chat/src/main/java/com/github/twitch4j/chat/util/MessageParser.java
```diff
@@ -129,4 +130,23 @@ private void parseTag(Map<String, CharSequence> tags, CharSequence tag, int deli
...
+                }
+                break;
+            } else {
+                boolean carriage = i > 0 && source.charAt(i - 1) == '\r';
+                consumer.accept(source.substring(start, carriage ? i - 1 : i));
+                start = i + 1;
+            }
+        }
+    }
+
```
PhilippHeuer, line 151: What do you think about this?

```suggestion
    @ApiStatus.Internal
    public void consumeLines(@NotNull String source, @NotNull Consumer<String> consumer) {
        int len = source.length();
        int start = 0;
        int i;

        while (start < len && (i = source.indexOf('\n', start)) >= 0) {
            boolean carriage = i > 0 && source.charAt(i - 1) == '\r';
            consumer.accept(source.substring(start, carriage ? i - 1 : i));
            start = i + 1;
        }

        if (start < len) {
            consumer.accept(source.substring(start));
        }
    }
```

iProdigy: Good refactor (I like removing the if/else) - I think we can remove `start < len && ` from the `while` condition (`indexOf` already has this check internally)

PhilippHeuer (approved):

PR #1102, opened 2025-01-30
fix: shutdown library-created executors on TwitchClientPool close

Awakened-Redstone:

### Changes Proposed

* Fixes thread name matching on `TwitchClientPool#close` to be the same of `TwitchClient#close`, 
* Fixes non daemon threads being left running


iProdigy (approved):
good catch!

